PR #2, opened 2022-09-12
draft for kepler CR

sunya-ch:
Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

### DIFF
--- a/doc/cr.md
+++ b/doc/cr.md
@@ -1,0 +1,56 @@
+## Kepler CR
+The Kepler CR provides specification of each kepler components including exportor, estimator, model-server, and exporter (including prometheus and grafana dependency.
+  
+```yaml
+collector:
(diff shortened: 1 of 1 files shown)


doc/cr.md
```diff
@@ -0,0 +1,56 @@
...
+    query-step:
+    sampling-interval:
+    enable-pipelines:
+    - * (all available pipelines)
+    - [pipeline name]
+    models-storage:
+      type: local
+      hostPath: 
+      ...
+prometheus:
```
rootfs, line 50: we can move the `prometheus` and `grafana` spec to another CR, e.g. `kepler-integration`, so that end users with different integration requirement can write their own CR and deploy kepler with other stacks.

PR #5, opened 2022-09-21
WIP Set up spec for dependencies and integrations

KaiyiLiu1234:
Updated spec for dependencies. I think for the integrations (Prometheus and Grafana) we will be using operators. We still need to set up a controller and cr for these integrations since we need to allow end users to customize their integrations. The controller reconcile can deploy the yaml or helm.


KaiyiLiu1234:
I was also wondering is there a specific kind of spec we need for the integration CRs, or can we just have the user input a CR that is accepted by the external integration operators we are using (eg. https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md)

api/v1/keplerintegration_types.go
```diff
@@ -0,0 +1,76 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// Grafana Spec
+
+type GrafanaSpec struct {
+	DashboardScript string `json:"dashboardScript,omitempty"`
```
rootfs, line 29: i would start with strong typed API to avoid poor formed input. Please consider import Grafana API pkg so we can define [GrafanaDataSource](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/openshift/dashboard/03-grafana-datasource-UPDATETHIS.yaml) with the API pkg [here](https://github.com/grafana-operator/grafana-operator/blob/v3.10.4/pkg/apis/integreatly/v1alpha1/grafanadatasource_types.go#L17), [Grafana instance](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/openshift/dashboard/02-grafana-instance.yaml) withe API pkg [here](https://pkg.go.dev/github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1#GrafanaSpec), etc

api/v1/keplerintegration_types.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Prometheus Spec
+
+type PrometheusSpec struct {
+}
```
rootfs, line 35: Prometheus is assumed running, define [ServiceMonitor](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/openshift/dashboard/00-servicemonitoring-kepler.yaml)

PROJECT
```diff
@@ -5,6 +5,9 @@ resources:
 - group: dependency
   kind: KeplerDependency
   version: v1
+- group: dependency
+  kind: KeplerIntegration
```
rootfs, line 9: maybe `KeplerGrafanIntegration` (in case there are other types of dashboard to integrate)

api/v1/keplerdependency_types.go
```diff
@@ -23,13 +23,94 @@ import (
...
+type CollectorSpec struct {
+	Image        string       `json:"image,omitempty"`
+	Sources      Sources      `json:"sources,omitempty"`
+	RatioMetrics RatioMetrics `json:"ratioMetrics,omitempty"`
+}
+
+// Estimater CRD Requirements
+
+type NodeSelector struct {
+	//not sure if this is correct
```
rootfs, line 68: you can use core [NodeSelectorTerms](https://pkg.go.dev/k8s.io/api/core/v1#NodeSelectorTerm) API to support more affinity rules

PROJECT
KaiyiLiu1234: Should we also create a KeplerPrometheusIntegration for ease of use? Currently KeplerIntegration houses a CR that covers both Grafana and Prometheus.

rootfs:
@KaiyiLiu1234 Thanks! This work is a very comprehensive workflow and API walkthrough

PROJECT
rootfs: sounds good!

KaiyiLiu1234:
The Operator has been changed to allow multiple groups. Two groups were added: Kepler System and Kepler Addons. Kepler System will contain controller(s) for Kepler, Model Server, and Estimator. Kepler Addons will contain controllers for MachineConfig and ServiceMonitor.

PR #22, opened 2022-12-13
Commit working unit test

KaiyiLiu1234:
This PR contains an example of a working unit test for ensureDaemonSet. The Unit tests will aim to cover all ensure functions and Reconcilers within CollectorReconciler. This should cover more than 50%. A Mock client has been set up to record basic Kubernetes API Client functionality and this will be used to determine whether the resources made are expected and contain the right fields. For testing reconcilers as a whole, we can test to see whether the desired resources have been created (as the ensure functions will be unit tested individually to ensure each resource is correct). Thus, I would like to also include TypeMeta Kind field when defining the resources to help distinguish between different resources.


KaiyiLiu1234:
Unit tests now easily cover 50-80% of working code. An issue I have encountered related to the Main Reconciler is that the Mock client Get does not update the obj struct pointer. Thus, when calling client Get in the main reconciler for the Kepler Instance, it will never work. I am not sure how to update the client.Object obj with the new fields as the interface seems to be lacking the needed methods. Any recommendations?

husky-parul:
@KaiyiLiu1234 can you rebase the PR to re run the tests?

KaiyiLiu1234:
Unit tests use kubernetes Fake Client for mocking kubernetes api. Currently, the Main Reconciler and all sub reconcilers are covered by unit tests.

husky-parul (approved):
LGTM

PR #31, opened 2023-01-13
MachineConfig and SCC support for OpenShift

husky-parul:
This PR add functionality to detect OpenShift env and skip MachineConfig and SCC if not OpenShift


KaiyiLiu1234 (approved):
PR includes up to date Bundle so the previous bundle PR does not need to be merged. For future, need to include a constant file for naming conventions of objects based on owner metadata name and some descriptive suffix. This should allow consistency. Note that machineconfig and scc controller don't seem to check if it is on openshift. I can include this change in a follow up PR since the code was already included in unit tests.

KaiyiLiu1234:
If new unit tests are included, we might have to update the bundle in upcoming PR

PR #52, opened 2023-03-03
Bugfix

husky-parul:
This PR includes:
    - [Bug fix](https://github.com/sustainable-computing-io/kepler-operator/issues/51) to export port from controller to crd
    - Updates README.MD and Contribution Guidelines
    - Provides manifests to deploy Kepler's default Grafana dashboard
    - Provides script for local kind cluster

PR #53, opened 2023-03-03
remove HostPort that causes conflicts with SNO's CNI over port 9102

chr15p:
With Single Node Openshift (and presuambly others) port 9102 is already used by the OVN Kubernetes CNI. Attempting to reuse it caused the kepler-exporter pod to get stuck in 'pending' because it couldn't get the hostport it wanted.  This commit removes the requirement for host networking and HostPort: 9102 from kepler-export, which allows it to run on.

This is the same issue as https://github.com/sustainable-computing-io/kepler/issues/527 and the same fix ported over.


rootfs:
also see #52 for port fixes in the crd and sample. The host network fix needs to be picked up.

rootfs:
@chr15p that makes sense. @husky-parul is it something can be done in this or next release? Thanks

promptparul:
We can take this in this release.

piparul (approved):
LGTM

PR #58, opened 2023-03-14
Update Model Server Unit Test for Debugging

KaiyiLiu1234:
Updated Model Server Unit Test. The update revealed some issues with the model server which will be addressed in followup commits


KaiyiLiu1234:
4 Commits have been squashed into 1 commit.

controllers/kepler_controller.go
```diff
@@ -429,16 +433,12 @@ func (r *collectorReconciler) ensureDaemonSet(l klog.Logger) (bool, error) {
...
-		image := r.Instance.Spec.Collector.Image
-		logger.V(1).Info("DaemonSet Image ", "image", image)
-
 		r.daemonSet.Spec.Template.Spec.Containers = []corev1.Container{{
 			Name: "kepler-exporter",
 			SecurityContext: &corev1.SecurityContext{
 				Privileged: &scc_value,
 			},
-			Image:   image,
+			Image:   "quay.io/sustainable_computing_io/kepler:latest",
```
piparul, line 441: Remove hard coding the image name from here

controllers/kepler_controller.go
```diff
@@ -429,16 +433,12 @@ func (r *collectorReconciler) ensureDaemonSet(l klog.Logger) (bool, error) {
 		r.daemonSet.Spec.Template.ObjectMeta.Name = dsName.Name
 		r.daemonSet.Spec.Template.Spec.HostNetwork = true
 		r.daemonSet.Spec.Template.Spec.ServiceAccountName = r.serviceAccount.Name
-
-		image := r.Instance.Spec.Collector.Image
```
piparul, line 433: Do not remove this line. We should read the image from the spec

piparul:
It seems your branch was not rebased with upstream `v1alpha1` as your commits have overridden the last commit made to v1alpha1. Can you ensure that the branch is rebased?

Also the PR needs some clean up. Lots of commented code. Can you clean the PR before we push it?

@KaiyiLiu1234

controllers/kepler_controller.go
```diff
@@ -161,56 +161,60 @@ func (r *KeplerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
-	/*
```
piparul, line 164: Can we get rid of the commented code?

controllers/kepler_controller.go
```diff
@@ -444,7 +440,10 @@ func (r *collectorReconciler) ensureDaemonSet(l klog.Logger) (bool, error) {
 				Name:          "http",
 			}},
 		}}
+		// Check if Estimator Patch is required
+		if r.Instance.Spec.Estimator != nil {
```
SamYuan1990, line 444: if possible, please double check lint issue.

config/samples/kepler.system_v1alpha1_kepler.yaml
```diff
@@ -8,10 +8,15 @@ metadata:
     app.kubernetes.io/managed-by: kustomize
     app.kubernetes.io/created-by: kepler-operator
   name: kepler
-  # namespace: kepler
 spec:
   # TODO(user): Add fields here
   collector:
-    image: quay.io/sustainable_computing_io/kepler:release-0.4
+    image: quay.io/sustainable_computing_io/kepler:latest
```
SamYuan1990, line 14: do we provide latest image as official release?

controllers/kepler_controller.go
```diff
@@ -161,56 +161,56 @@ func (r *KeplerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-		// Add finalizer for this CR
-		if !ctrlutil.ContainsFinalizer(inst, keplerFinalizer) {
-			ctrlutil.AddFinalizer(inst, keplerFinalizer)
-			updateErr := r.Client.Update(ctx, inst)
-			if updateErr != nil {
-				return ctrl.Result{}, updateErr
-			}
-		}
-	*/
+	//if inst set to delete, pv must also be removed when it is released
```
SamYuan1990, line 165: just remove them for clean?

SamYuan1990 (commented):
I am glad to see test case update. Before we merge it, I leave some comments.

config/samples/kepler.system_v1alpha1_kepler.yaml
KaiyiLiu1234: I believe we do. @piparul can you confirm? This is just a sample of how the CRD would look like for the kepler operator so any image version theoretically can be passed to the operator when deploying kepler or kepler-model-server.

controllers/kepler_controller.go
KaiyiLiu1234: Right. I believe this will be needed for including the estimator patch to kepler's daemonset. I will remove this from this PR because @piparul will be implementing the estimator patch and this is incomplete.

controllers/kepler_controller.go
KaiyiLiu1234: I will remove this. But the reason this exists is that if the CRD is removed, all resources will be terminated automatically except for the PV and PVC. These lines will include a finalizer to remove the PV and PVC. However, I think it would be smarter if we just left this for the user to decide to remove the PV and PVC. We should instead include a script that could automatically remove the PV and PVC that is triggered by the user and will only activate if the CR does not exist. Thoughts?

KaiyiLiu1234: Or we can include a field (RemovePVUponDeletion) in the CRD which will be a boolean and if set to true will add the finalizer. Or we can completely ignore it and the PV/PVC will remain even if CR is deleted and it is up to user to address it.

config/samples/kepler.system_v1alpha1_kepler.yaml
piparul: Yes we do. Whenever we cut a release, we provide an official release image. https://quay.io/repository/sustainable_computing_io/kepler-operator?tab=tags

controllers/kepler_controller.go
piparul: >We should instead include a script that could automatically remove the PV and PVC that is triggered by the user and will only activate if the CR does not exist. Thoughts?
@KaiyiLiu1234 Maintaining script that is not part of the operator code/pkg doesn't seems right to me. 
@SamYuan1990 Could you elaborate on your thoughts as to why we should not have the finalizer to cleanup PVCs?

KaiyiLiu1234: Actually this is a very minor change since the code for including the finalizer is there already. I will make the change here

KaiyiLiu1234:
Finalizer has been included. I have encountered a unit test error stating the finalizer was not added. Will test further and repush today.

KaiyiLiu1234:
Unit tests now cover 75.1%. Finalizer unit test was added but it is a bit rudimentary (just checks if the correct finalizer was added to the kepler instance - we can discover the finalizers added by looking at keplerInstance.GetFinalizers()). I wanted to try and invoke deletion request of the kepler instance (by using r.client.delete) to see if it triggers the garbage collection to deal with finalizers and the other kubernetes resources but interestingly it does not. Any ideas if I am doing something wrong?

SamYuan1990 (approved):
LGTM

controllers/kepler_controller.go
SamYuan1990: Sorry I just found I am been mentioned in a reference...
@piparul , @KaiyiLiu1234 my original comments is that we should avoid use comments as `/**/` but we should remove pv/pvc once the instance been removed.

piparul:
> Unit tests now cover 75.1%. Finalizer unit test was added but it is a bit rudimentary (just checks if the correct finalizer was added to the kepler instance - we can discover the finalizers added by looking at keplerInstance.GetFinalizers()). I wanted to try and invoke deletion request of the kepler instance (by using r.client.delete) to see if it triggers the garbage collection to deal with finalizers and the other kubernetes resources but interestingly it does not. Any ideas if I am doing something wrong?

That's interesting. I will have to look into this. We can merge the PR for now. @KaiyiLiu1234 can you open an issue to address this down the line?

piparul (approved):
LGTM

PR #85, opened 2023-07-07
bump CI version to v0.0.1

SamYuan1990:


rootfs (approved):

SamYuan1990:
I suppose we need update script to upload images to mircoshift and the change can be ref to kepler's CI. :-)  I will take a look and update later.

SamYuan1990:
Hi @rootfs , I suppose for operator test, we need do further updates for microshift. please review this PR.

rootfs:
@husky-parul can you take a look?

vprashar2929:
> Hi @rootfs , I suppose for operator test, we need do further updates for microshift. please review this PR.

Yes, absolutely I agree. I was planning to integrate microshift for Operator but I am happy to add further updates once this PR is merged.

vprashar2929:
@SamYuan1990 I was thinking can we also add creating an instance of Kepler once the Operator is deployed since it will also allow us to validate things e2e for instance there can be one scenario where there is any pre-req that is supposed to be handled by the Operator installation but didn't happen and when Kepler instance is launched it fails?

SamYuan1990:
> @SamYuan1990 I was thinking can we also add creating an instance of Kepler once the Operator is deployed since it will also allow us to validate things e2e for instance there can be one scenario where there is any pre-req that is supposed to be handled by the Operator installation but didn't happen and when Kepler instance is launched it fails?

- For kepler repo, it's run e2e test case and it's cover it's own test case for kepler pod.
- For kepler Operator, it's running as an operator as a pod.
- For this PR, it's about CI as kepler-action bump up.
- For kelper Operator,  I suppose it 's better to as @rootfs  and @piparul for this question as it's out of this PR's scope.
> which the question is: for any operator release with kepler release, will we have/add a test case which 1st deploy kepler pod with specific version, and then run e2e test case download from kepler(with specific version) repo.

PR #88, opened 2023-07-11
Remove machine config from operator

husky-parul:
This PR removes machine config dependencies from the operator


go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/onsi/ginkgo/v2 v2.6.0
 	github.com/onsi/gomega v1.24.1
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
```
sthaha, line 9: Do we need this now?

go.mod
```diff
@@ -7,7 +7,6 @@ require (
...
-	github.com/openshift/machine-config-operator v0.0.1-0.20200913004441-7eba765c69c9
```
sthaha, line 10: Could you please run `go.mod tidy`? 

It may be worth adding a lint for that WDTY?

controllers/kepler_controller.go
```diff
@@ -88,7 +88,6 @@ func (r *KeplerReconciler) removePVC(logger logr.Logger, inst *keplerv1alpha1.Ke
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=*,verbs=*
 //+kubebuilder:rbac:groups=core,resources=nodes/metrics;nodes/proxy;nodes/stats,verbs=get;list;watch
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch;create;update;patch;delete;use
-//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs;machineconfigpools,verbs=get;list;watch;create;update;patch;delete;use
```
sthaha, line 91: Shouldn't we `make bundle` and update the bundle as well?

sthaha (commented):
I tested this running the operator out of cluster against openshift 4.13-nightly and querying kepler exporter service for metrics which seems to work fine. 

I think we should ensure
-  the bundle is up to date 
-  go mod tidy is done as well 

On an different topic, why are we vendoring (go mod vendor) instead of relying on `go mod` itself to resolve the dependencies ?

husky-parul:
> the bundle is up to date

We already have the steps to make/validate and test bundle in the Makefile. I want to automate the bundle part as a GHA post PR merge into this [repo](https://github.com/sustainable-computing-io/community-operators-prod) through which we raise PR in the community catalog. Maybe we can open an issue and discuss this pipeline? 


> go mod tidy is done as well

Done

> On an different topic, why are we vendoring (go mod vendor) instead of relying on go mod itself to resolve the dependencies ?

I will update the `.gitignore` for `vendor`

sthaha:
/lgtm , please feel free to merge @husky-parul 

@vprashar2929, just FYI something to test on Openshift :)

PR #89, opened 2023-07-12
Improve running operator locally

sthaha:
This PR improves on how operator is run locally by
  1. installing local-dev-cluster to a tmp directory
  2. moving all tools (controller-gen, kustomize etc) to tmp/bin
  3. Adding a section explaining running operator out of cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 ##@ Build Dependencies
 
 ## Location to install dependencies to
-LOCALBIN ?= $(shell pwd)/bin
+LOCALBIN ?= $(shell pwd)/tmp/bin
 $(LOCALBIN):
 	mkdir -p $(LOCALBIN)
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -266,28 +266,34 @@ set_govulncheck:
...
 prepareKubeConfig:
 	./hack/prepareKubeConfig.sh
-.PHONY: prepareKubeConfig
 
-cluster-up:
-	rm -rf local-dev-cluster
-	git clone -b v0.0.1 https://github.com/sustainable-computing-io/local-dev-cluster.git --depth=1
-	cd local-dev-cluster && ./main.sh
 .PHONY: cluster-up
+cluster-up:
```
vprashar2929, line 286: Can we move the code to a separate script inside the hack folder? Like we have in [Kepler](https://github.com/sustainable-computing-io/kepler/blob/main/hack/cluster-up.sh) repo

vprashar2929 (dismissed):

Makefile
sthaha: yes, but beyond the scope of this PR :)

husky-parul (approved):
lgtm

README.md
```diff
@@ -16,10 +16,20 @@ You’ll need a Kubernetes cluster to run against. You can use [KIND](https://si
 ### To run a kind cluster locally 
 
 ```sh
-make cluster-up CLUSTER_PROVIDER='kind' CI_DEPLOY=true
+make cluster-up
```
vprashar2929, line 19: Since `local-dev-cluster` also supports microshift I think it's best to have `CLUSTER_PROVIDER` env variable

README.md
```diff
@@ -16,10 +16,20 @@ You’ll need a Kubernetes cluster to run against. You can use [KIND](https://si
...
+kind get kubeconfig > tmp/kubeconfig
```
vprashar2929, line 20: Same here as well in case we want to export the kubeconfig for microshift

sthaha: sure but the point is that as it currently stands, defining these here have no value, so beyond the scope of this pr :)

Makefile
sthaha: @vprashar2929 please feel free to open PR since I think there is value in keeping all non-trivial scripts in `/hack`

README.md
sthaha: @vprashar2929 I feel the suggestion is valuable, could you please open a PR for this so that these env vars are  honoured ?

vprashar2929: Sure sounds good. I will open a PR related to this.

PR #91, opened 2023-07-13
Fix make deploy

vprashar2929:
This PR fixes `make deploy` when used to deploy the operator on local env.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,9 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	@if [ "$(IMG)" != "controller:latest" ];then \
+		cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG); \
+	fi
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 .PHONY: undeploy
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/sustainable_computing_io/kepler-operator
-  newTag: "0.5"
+  newTag: "0.5.0"


Makefile
```diff
@@ -159,6 +159,9 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	@if [ "$(IMG)" != "controller:latest" ];then \
+		cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG); \
```
husky-parul, line 163: For me using `make deploy IMG=<>` kustomize automatically updates the controller. Why we need this?

vprashar2929: strange🤔 I don't see this on my local machine. How is it automatically updating when we didn't add the step to edit the image?

For me, it just deploys Operator with [this](https://github.com/sustainable-computing-io/kepler-operator/blob/afb499daebde69e576f858e5dbe72662bc97dd14/config/manager/kustomization.yaml#L8). Currently, we have `0.5` tag which is incorrect and the pod starts to show `ErrpullImage` that's when I realized we are not changing the image actually.

Makefile
```diff
@@ -159,6 +159,9 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	@if [ "$(IMG)" != "controller:latest" ];then \
```
sthaha, line 162: @vprashar2929 note for future use `[[` instead of `[` :) . Also why is this check needed? why not just cd and set ?

vprashar2929: I also had the same idea to do just cd but I noticed that since we already mantain the operator image with tag. It's best to have the flexibility to make setting the image optional. If anyone wants to use the latest image then can provide that in argument and if not it will pick up whatever image tag is present [here](https://github.com/sustainable-computing-io/kepler-operator/blob/c8e36ad8885e7705a2175756f51d75d97f969f33/config/manager/kustomization.yaml#L8).

sthaha: IMG = controller:latest isn't valid image so,  this will create a broken deployment anyway. I don't think `controller:latest` should be treated as somehow special. I will provide a PR that fixes this or please feel free to do that :)

PR #97, opened 2023-07-18
Add support for podman

vprashar2929:
This PR:
* Add's support for using Podman as the container runtime.
* Fixes the IMG variable value.
* Update GH workflows wherever `make docker-` was used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.4.0
+VERSION ?= 0.5.0
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
@@ -23,6 +23,13 @@ ifneq ($(origin DEFAULT_CHANNEL), undefined)
 BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
+
+# CTR_CMD defines the container runtime environment to be used when building and pushing images.
+ifdef CTR_CMD
+	CTR_CMD := $(CTR_CMD)
+else
+	CTR_CMD :=$(or $(shell which podman 2>/dev/null), $(shell which docker 2>/dev/null))
+endif
 
 # IMAGE_TAG_BASE defines the docker.io namespace and part of the image name for remote images.
 # This variable is used to construct full image tags for bundle and catalog images.
@@ -47,7 +54,7 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+OPERATOR_IMG ?= $(IMAGE_TAG_BASE):latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.25.0
 
@@ -62,6 +69,10 @@ endif
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
+
+# Setting GOENV
+GOOS := $(shell go env GOOS)
+GOARCH := $(shell go env GOARCH)
 
 .PHONY: all
 all: build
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -134,14 +140,14 @@ docker-push: ## Push docker image with the manager.
...
-docker-buildx: test ## Build and push docker image for the manager for cross-platform support
+.PHONY: operator-buildx
+operator-buildx: test ## Build and push docker image for the manager for cross-platform support
 	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
 	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
-	- docker buildx create --name project-v3-builder
-	docker buildx use project-v3-builder
-	- docker buildx build --push --platform=$(PLATFORMS) --tag ${IMG} -f Dockerfile.cross
-	- docker buildx rm project-v3-builder
+	- $(CTR_CMD) buildx create --name project-v3-builder
```
husky-parul, line 147: Can you confirm if podman support `buildx`? Last time I checked only `buildah` had cross platform build compatibilty.
[Related issue in podman ](https://github.com/containers/buildah/issues/1590)

vprashar2929: I think we don't actually need this as podman and docker both supports `--platform` flag. I can update the `make operator-build` step to have a platform flag added. Wdyt?

husky-parul: Yes we need to make sure that the operator bundles are created and validated correctly for podman too.

Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.4.0
+VERSION ?= 0.5.0
```
sthaha, line 6: Unrelated to this PR

A good idea would be to keep `VERSION` file and read from it. The file would be the source of truth for all tools that need to know the version.

Makefile
```diff
@@ -47,7 +54,7 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= $(IMAGE_TAG_BASE):latest
```
sthaha, line 57: while at it, could you please rename this to OPERATOR_IMG ? There will be other images as well. I am especially thinking of publishing bundles which takes us one step closer to having (or not having) a catalog.

Makefile
```diff
@@ -115,34 +126,18 @@ build: generate fmt vet ## Build manager binary.
...
-	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
-	- docker buildx create --name project-v3-builder
-	docker buildx use project-v3-builder
-	- docker buildx build --push --platform=$(PLATFORMS) --tag ${IMG} -f Dockerfile.cross
-	- docker buildx rm project-v3-builder
-	rm Dockerfile.cross
+	$(CTR_CMD) build -t ${IMG} \
+	--build-arg TARGETOS=$(GOOS) \
+	--build-arg TARGETARCH=$(GOARCH) \
+	--platform=linux/$(GOARCH) .
```
sthaha, line 136: Does this help create multi-arch images? I haven't played around with it but my reference would be - https://github.com/prometheus-operator/prometheus-operator/blob/main/scripts/push-docker-image.sh

sthaha: Could you please point to a multi-arch image created using this?

sthaha (changes requested):

Makefile
vprashar2929: > Does this help create multi-arch images? 

Yes, I tried both with podman as well as docker. If do a inspect on the image you will be able to see the architecture which you specified in the platform flag.

> Could you please point to a multi-arch image created using this?

[Kepler](https://github.com/sustainable-computing-io/kepler/blob/f94a6651894f875f556f7494934055dc681ad8ce/Makefile#L179) repo uses this to build multi-arch images

sthaha: I will test this when I have a bit more time, meanwhile, could you please fix the GitHub action as well?

vprashar2929: Do we want CI to build operator image using docker or podman? Since GH runner supports both it uses podman by default.
Also, we have steps in `integration_test` workflow file and `prepareKubeConfig.sh` which uses docker, and that will require an update accordingly

vprashar2929:
@SamYuan1990 Thanks for sharing your investigation. Since I can see [this](https://github.com/kubernetes-sigs/kind/issues/3261) is still an open issue I think we can still go with docker when it comes to deploying kind but for building Operator image we can use podman. Wdyt?

SamYuan1990:
> @SamYuan1990 Thanks for sharing your investigation. Since I can see [this](https://github.com/kubernetes-sigs/kind/issues/3261) is still an open issue I think we can still go with docker when it comes to deploying kind but for building Operator image we can use podman. Wdyt?

Honestly, 
1st at last time I took a look at podman, which seems just a CLI following open container run time. Hence, I am curious on which open container run time we are going to use in this issue. instead a command line tool as podman...
2nd as either docker or runc, containerd etc those open container run time follow open container specs... hence may I know why should we .... worried about docker?(in other words, either of container run time are good for us, because we don't use any tooling specific feature during build or test?)
specific for podman
even if we ref this https://github.com/kubernetes-sigs/kind/blob/main/.github/workflows/podman.yml, we can't get https://github.com/kubernetes-sigs/kind/blob/main/.github/workflows/cgroup2.yaml support on a linux GHA agent, and have to invoke mac and build a VM on mac... etc etc.... in my point of view, the effort to support the cgroupv2 and related items are too large/huge than we received.

Hence when I review the start point of this issue, as podman is just a CLI, but not container run time, hence why? https://github.com/containers/podman#podman-a-tool-for-managing-oci-containers-and-pods

sthaha:
@vprashar2929 do we want to continue this path? IMHO, it is best to standardise on just one tool unless the other option gives us a huge benefit at which point, we should consider switching. I have no strong opinion on which tool to use but lets pick one and stick with it.  👼

sthaha:
@vprashar2929 , I am closing this PR since this needs to be rebased. Please reopen the PR if you disagree

Issue #99, opened 2023-07-19
Make Kepler CRD cluster-scoped and Unique


sthaha:
There can only be a single instance of kepler per cluster, it makes sense to make the `Kepler` CRD cluster scoped. This breaks the API and allows us to refine the current API/ Spec. 

I am working on [enhancement proposal here](https://docs.google.com/document/d/1vBRLdGFbfU-1tZ7DQwo4a3Hhv0QwzPXA8LbO73Bi--M/edit). Please feel free to collaborate / co-author.

Issue #90, opened 2023-07-12
Uninstall Kepler doesn't remove clusterroles, rolebindings and scc on OpenShift


vprashar2929:
When the Kepler instance is created it creates the appropriate cluster role, role bindings, SCC, and service accounts required by the exporter to run. When we uninstall the instance it removes the SA from the deployed namespace but doesn't remove cluster roles, cluster role bindings, and SCC. The removal of these also doesn't happen when we uninstall the Kepler operator itself from the cluster.

Steps to reproduce:
* Deploy Kepler Operator via Operator hub on OpenShift.
* Create a Kepler Instance(It can be inside the `openshift-operators` namespace or different)
* Wait for daemonset's to come up.
* Check that the cluster roles, cluster role bindings, service account, and securitycontextconstraints are created.
* Delete the Kepler Instance/Operator from the console UI.

I think the removal of cluster roles, cluster role bindings, service account, and securitycontextconstraints should be handled by the Operator

husky-parul:
@vprashar2929 this is strange. Because `namespaced` based resources gets deleted when we uninstall the operator. We've faced similar issues with `PVCs` in the past and the fix was to include the PVCs in the [kepler finalizers. ](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/controllers/kepler_controller.go#L117)
Maybe we can include the `scc, cluster roles, cluster role bindings and service account` too. 

Before that can we confirm if this is an expected behavior or that we have missed some `operator-sdk` implementation post operator uninstall.

@sthaha any idea?

sthaha:
> When the Kepler instance is created it creates the appropriate cluster role, role bindings, SCC, and service accounts required by the exporter to run. When we uninstall the instance it removes the SA from the deployed namespace but doesn't remove cluster roles, cluster role bindings, and SCC.

This is a bug which can fixed by using `ownerReferences` 
@husky-parul ☝️

>  The removal of these also doesn't happen when we uninstall the Kepler operator itself from the cluster.

This isn't a bug and is the expected behaviour when **any** operator is uninstalled. 
@vprashar2929

sthaha:
@husky-parul could you please shed some light on the use of finalizers ? 

> . Because namespaced based resources gets deleted when we uninstall the operator 

Uninstalling the operator should not delete the kepler CR and AFAIK should not affect any running kepler instances. 
Based in my limited knowledge of the operator, I don't think we need finalizers . 
[...]

Issue #84, opened 2023-07-07
code improvement - use SSA instead of Get / Update / Create


sthaha:
All of the code that updates a resource can benefit from [using SSA which also allows us](https://kubernetes.io/docs/reference/using-api/server-side-apply/) to specify the `FieldManagers` allowing external controllers to modify spec of operands that kepler-operators doesn't really need to own (e.g. resources, limits etc). Most CRUD code will simplify to 

```go
	// Apply the changes using Server-Side Apply
	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
		_, err := r.client.Apply(context.TODO(), obj, metav1.ApplyOptions{
			FieldManager: "kepler-operator",
			Force:        true,
		})
		return err
	})
        // do something about retryErr or 
	return retryErr
```

Issue #82, opened 2023-07-07
Add ObservedGeneration to status.Conditons


sthaha:
ObservedGenerations operators (including human operators) to react to changes only if `kepler.metadata.Generation == kepler.Status.<Interested-Condition>.ObservedGeneration`

I.E. It allows external operators to make a change to the `kepler.Spec` which (automatically) modifies `kepler.metadata.Generation`, watch the object for changes, and only react if the condition (say Available) is `True` and the `ObservedGeneration` == `Generation` 

Additional note based on #83 

The ObservedGeneration field is typically updated during the reconciliation process based on specific events or conditions. For the `Reconcile` condition, it is often updated before the reconciliation process starts. On the other hand, for the `Available` condition, it is usually updated after the reconciliation process is completed and the availability status of the resource is determined.

PR #100, opened 2023-07-21
Refactor kepler implementation

sthaha:
**NOTE:** This breaks API, fixes bugs and adds missing feature. Most Notably 

- Makes Kepler CRD clusterscoped 
- Adds `conditons` List in Status instead of a single condition, so that `Available` conditions can be added as followup
- Reconciled Condition now has ObservedGeneration and sets LastTransitionTime correctly
- Ensures only a single instance of kepler named kepler will be reconciled  
- Cleanup resources correctly on deletion of kepler
-  Deploys kepler exporter to openshift-kepler-operator namespace
- Rearchitects reconciler pattern
- Upgrade controller-gen to 0.12.1 (subsequently `go` to 1.20)

Pending
- Unit tests (in a follow-up PR)

Fixes: 
* https://github.com/sustainable-computing-io/kepler-operator/issues/99
* https://github.com/sustainable-computing-io/kepler-operator/issues/90
* https://github.com/sustainable-computing-io/kepler-operator/issues/84
* https://github.com/sustainable-computing-io/kepler-operator/issues/82


acornett21:
Would it be possible for the removal of the vendor folder to be in a different PR? Having this all in one PR makes this PR not renewable in the GH UI due to their 3k change load limit.

.github/workflows/gofmt.yml
```diff
@@ -10,6 +10,6 @@ jobs:
       - uses: actions/checkout@main
       - uses: actions/setup-go@main
         with:
-          go-version: 1.19
+          go-version: "1.20"
```
acornett21, line 13: It might be nice to use the `v4` version of setup-go so the builds can reference the mod file and the actions need not be updated. Snippet below
```
      uses: actions/setup-go@v4
      with:
        go-version-file: go.mod
```

Dockerfile
```diff
@@ -12,16 +12,17 @@ COPY go.sum go.sum
...
+
 
 # Build
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} \
+  go build -a -o manager ./cmd/manager/...
```
acornett21, line 25: Would it make more sense here to update the make file with the proper build args and just call `make build` here, and have the proper args in the make file? This would ensure that the binary built locally/ci and in the docker file match, and would reduce the places that would need to change if it's determined that the binary needs to be built differently, ie if build flags for version/commit/etc were to be added alter down the road.

sthaha: That's a great suggestion, would you be okay with filing an issue and we can handle that separately ?

.github/workflows/gofmt.yml
sthaha: I have filed - https://github.com/sustainable-computing-io/kepler-operator/issues/101

husky-parul:
@sthaha I don't see Kepler -exporter pod when deploying from the refactor branch

Steps to reproduce:

- set up kind cluster using `make cluster-up`
- checkout the `refactor-kepler` branch
- `make docker-build docker-push IMG=quay.io/husky_parul/kepler-operator:refactor`
- `make deploy IMG=quay.io/husky_parul/kepler-operator:refactor`
- `oc apply -k config/sample`
- `oc get pods --all-namespaces | grep kepler`
```
kepler-operator-system   kepler-operator-controller-manager-795c5d964d-djfgv   2/2     Running   0          7m53s

oc get kepler
NAME     AGE
kepler   8m2s

oc get daemonset --all-namespaces
NAMESPACE     NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
kube-system   kindnet          1         1         1       1            1           <none>                   53m
kube-system   kube-proxy    1         1         1       1            1           kubernetes.io/os=linux   53m



```

sthaha:
> I don't see Kepler -exporter pod when deploying from the refactor branch. No DS are created. Kepler CR is present. The operator manager container logs are not showing any updates for reconciliation as well.

I haven't tested this against OpenShift. I think it is most likely some rbac issue. Let me test it as well and get back.  BTW, a better way to deploy is to create the bundle and  use `operator-sdk run bundle`

vimalk78:
I used same steps and i am getting podsecurity error

```
  Warning  FailedCreate  6m32s (x10 over 17m)  daemonset-controller  (combined from similar events): Error creating: pods "kepler-exporter-ds-d6f28" is forbidden: violates PodSecurity "restricted:latest": privileged (container "kepler-exporter" must not set securityContext.privileged=true), allowPrivilegeEscalation != false (container "kepler-exporter" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kepler-exporter" must set securityContext.capabilities.drop=["ALL"]; container "kepler-exporter" must not include "SYS_ADMIN" in securityContext.capabilities.add), restricted volume types (volumes "lib-modules", "tracing", "proc", "kernel-debug", "kernel-src" use restricted volume type "hostPath"), runAsNonRoot != true (pod or container "kepler-exporter" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "kepler-exporter" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")

```

sthaha:
@husky-parul , @vimalk78 has found another issue with SCC running on Openshift. Please hold the review until that is fixed as well.

husky-parul:
>I haven't tested this against OpenShift. 
@sthaha  FYI I tested on a **kind** cluster

sthaha:
@husky-parul , @vimalk78 I tested the same on 4.13-nightly and the operator worked fine. The deployment model I used 
was to rely on `operator-sdk run bundle`

1. `make docker-build docker-push IMG=quay.io/sthaha/kepler-operator:refactor`
2. ` make bundle bundle-build bundle-push IMG=quay.io/sthaha/kepler-operator:refactor BUNDLE_IMG=quay.io/sthaha/kepler-operator-bundle:refactor`
3. ` operator-sdk run bundle quay.io/sthaha/kepler-operator-bundle:refactor --install-mode AllNamespaces --namespace=openshift-operators`
```
operator-sdk run bundle quay.io/sthaha/kepler-operator-bundle:refactor --install-mode AllNamespaces --namespace=openshift-operators                                           ⇡ remove-vendor 2d ● ⬡
INFO[0016] Creating a File-Based Catalog of the bundle "quay.io/sthaha/kepler-operator-bundle:refactor"
INFO[0019] Generated a valid File-Based Catalog
INFO[0024] Created registry pod: quay-io-sthaha-kepler-operator-bundle-refactor
INFO[0054]   Found ClusterServiceVersion "openshift-operators/kepler-operator.v0.4.0" phase: Succeeded
INFO[0054] OLM has successfully installed "kepler-operator.v0.4.0"
```
4. Ensure the operator image is right 
```
kubectl get deployment.apps/kepler-operator-controller-manager -o yaml -n openshift-operators | grep image:  
        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
        image: quay.io/sthaha/kepler-operator:refactor
```
5. ` kubectl logs deployment.apps/kepler-operator-controller-manager -f -n openshift-operators`
6.  `kubectl apply -k config/sample`
7.  `kubectl port-forward -n openshift-kepler-operator --address 0.0.0.0 svc/kepler-exporter-svc 9103:9103`
8.   `curl http://localhost:9103/metrics | grep 'joules'`

pkg/components/exporter/exporter.go
```diff
@@ -0,0 +1,341 @@
...
+
+const (
+	prefix = "kepler-exporter-"
+
+	SCCName = prefix + "scc"
+
+	ServiceAccountName   = prefix + "sa"
+	FQServiceAccountName = "system:serviceaccount:" + components.Namespace + ":" + ServiceAccountName
+
+	ClusterRoleName        = prefix + "cr"
```
vprashar2929, line 28: Can we make it ClusterRoleName = prefix + "clusterrole"?

pkg/components/exporter/exporter.go
```diff
@@ -0,0 +1,341 @@
...
+	ClusterRoleBindingName = prefix + "crb"
```
vprashar2929, line 29: Same here ClusterRoleBindingName = prefix + "clusterrole-binding"?

pkg/components/exporter/exporter.go
```diff
@@ -0,0 +1,341 @@
...
+
+	ConfigmapName = prefix + "cm"
+	DaemonSetName = prefix + "ds"
```
vprashar2929, line 32: Same here DaemonSetName = prefix + "daemonset"?

pkg/components/exporter/exporter.go
```diff
@@ -0,0 +1,341 @@
...
+	ServiceName   = prefix + "svc"
```
vprashar2929, line 33: Here as well ServiceName = prefix + "service"?

pkg/controllers/kepler.go
```diff
@@ -0,0 +1,300 @@
...
+
+	})
+}
+
+func (r KeplerReconciler) reconcilersForKepler(k *v1alpha1.Kepler) []reconciler.Reconciler {
+	// start with namespace
+	rs := []reconciler.Reconciler{}
+
+	if cleanup := !k.DeletionTimestamp.IsZero(); cleanup {
+		rs = append(rs, deletersForResources(components.NewKeplerNamespace())...)
```
vprashar2929, line 199: Won't this delete Operator as well from Namespace?

pkg/components/exporter/exporter.go
sthaha: Actually I was thinking the other way around to remove the suffix completely because `kind` is already a good enough differentiator.

pkg/controllers/kepler.go
sthaha: No, it wont .. Operator gets installed to `openshift-operators` and kepler-operator creates another namespace for deploying its resources which what is being deleted here.

Dockerfile
sthaha: Thank you 🙇 . I am resolving this conversation

pkg/components/exporter/exporter.go
vprashar2929: removing suffix is also fine but if we go with it then I suggest we should mention it completely.

pkg/components/exporter/exporter.go
```diff
@@ -0,0 +1,341 @@
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ConfigmapName,
+			Namespace: components.Namespace,
+			Labels:    labels,
+		},
+		Data: map[string]string{
```
rootfs, line 160: @husky-parul has made some changes to the configmap recently, and some defaults are also updated in [kepler repo](https://github.com/sustainable-computing-io/kepler/blob/main/manifests/config/exporter/exporter.yaml#L16)

KaiyiLiu1234 (approved):

hack/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022.
+Copyright 2023.
```
rootfs, line 2: please also add the boilerplate to the go files

config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
```diff
@@ -81,91 +84,85 @@ spec:
...
-                properties:
-                  includePVandPVCFinalizer:
-                    default: false
-                    type: boolean
                 type: object
             type: object
           status:
             description: KeplerStatus defines the observed state of Kepler
             properties:
               conditions:
```
rootfs, line 95: Is it a good time to make `conditions` an array to align with existing practices?

config/rbac/role.yaml
```diff
@@ -34,6 +34,18 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
rootfs, line 37: Does the operator need all the API access?

config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
husky-parul: This refactor will include PI changes so the plan is to include all changes to API in this release. Also the condition should be an array type as per the specification. This was the reason why the console was breaking because it expected an array type and was receiving an object.

husky-parul:
This refactor will include API changes so the plan is to include all changes to API in this release. Please give a keen review to the file pkg/api/v1alpha1/kepler_types.go
https://github.com/sustainable-computing-io/kepler-operator/blob/6e0a00d2f3bff7d215852e98aa603e3acae7829c/pkg/api/v1alpha1/kepler_types.go

@rootfs @KaiyiLiu1234

pkg/reconciler/finalizer.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+type Finalizer struct {
```
KaiyiLiu1234, line 14: Is this finalizer file being used at all during cleanup after Kepler resource has been deleted? From what I have seen, it seems only the deleter.go is being used. Originally, finalizer was only being used for cleanup of resources connected to Kepler Resources.

KaiyiLiu1234 (changes requested):

config/rbac/role.yaml
husky-parul: The core api includes 
```
kubectl api-resources --api-group=''
NAME                     SHORTNAMES   APIVERSION   NAMESPACED   KIND
bindings                              v1           true         Binding
componentstatuses        cs           v1           false        ComponentStatus
configmaps               cm           v1           true         ConfigMap
endpoints                ep           v1           true         Endpoints
events                   ev           v1           true         Event
limitranges              limits       v1           true         LimitRange
namespaces               ns           v1           false        Namespace
nodes                    no           v1           false        Node
persistentvolumeclaims   pvc          v1           true         PersistentVolumeClaim
persistentvolumes        pv           v1           false        PersistentVolume
pods                     po           v1           true         Pod
podtemplates                          v1           true         PodTemplate
replicationcontrollers   rc           v1           true         ReplicationController
resourcequotas           quota        v1           true         ResourceQuota
secrets                               v1           true         Secret
serviceaccounts          sa           v1           true         ServiceAccount
services                 svc          v1           true         Service
```
AFAIK we are using all the core api group (PVC/PV will come down the line when model server is incorporate)
@sthaha please confirm

pkg/reconciler/finalizer.go
KaiyiLiu1234: > Is this finalizer file being used at all during cleanup after Kepler resource has been deleted? From what I have seen, it seems only the deleter.go is being used. Originally, finalizer was only being used for cleanup of resources connected to Kepler Resources.

KaiyiLiu1234:
> This refactor will include API changes so the plan is to include all changes to API in this release. Please give a keen review to the file pkg/api/v1alpha1/kepler_types.go https://github.com/sustainable-computing-io/kepler-operator/blob/6e0a00d2f3bff7d215852e98aa603e3acae7829c/pkg/api/v1alpha1/kepler_types.go
> 
> @rootfs @KaiyiLiu1234

@husky-parul  @rootfs  While I think it is ok, I want to know your perspective on having the trainer spec present in the exporter spec instead of being present in the Kepler spec alongside the exporter spec. The original idea is that if the Model server is ever deployed, the exporter will always be deployed but not necessarily the trainer and the trainer will never be deployed without the exporter. Would there be situations in future where we only want the trainer deployed without the exporter?

husky-parul:
@KaiyiLiu1234 The `ModelServerExporterSpec` has been renamed to `ModelServerSpec`, rest is the same.
Other than `ModelServerTrainerSpec` is still part of `ModelServerSpec`

[current v1alpha1
](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/api/v1alpha1/kepler_types.go)
[refactor](https://github.com/sustainable-computing-io/kepler-operator/blob/6e0a00d2f3bff7d215852e98aa603e3acae7829c/pkg/api/v1alpha1/kepler_types.go#L85C11-L85C33)

husky-parul:
I see that we have removed `[Estimator](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/api/v1alpha1/kepler_types.go#L112)` from Kepler spec. Although it is not part of Dev Preview, incorporating it again later would need API changes.

config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
sthaha: Indeed it is, as we are breaking the API anyway.

config/rbac/role.yaml
sthaha: Yes, we will need access to PVC but do  note this PR does not  address the RBAC used. I have kept it the same. IIUC, we shouldn't need `get` since `list` and `watch` is what controller-runtime uses to build its cache. This is a minor detail we can address later.

pkg/components/exporter/exporter.go
sthaha: @rootfs shall I address this is separate PR?  - filed 
https://github.com/sustainable-computing-io/kepler-operator/issues/111

pkg/reconciler/finalizer.go
sthaha: Yes, we are using finalizer to ensure all cluster-scoped resources are deleted before deleting kepler

sthaha:
> I see that we have removed [Estimator](https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/api/v1alpha1/kepler_types.go#L112) from Kepler spec. Although it is not part of Dev Preview, incorporating it again later would need API changes.

@husky-parul New Api added to a CRD is not considered an API breakage, hence it is best to curb down on the API we expose.

hack/boilerplate.go.txt
sthaha: done :)

husky-parul (approved):
LGTM. We will take the reviews in the follow up PRs

sthaha:
@KaiyiLiu1234 , for the dev-preview we wouldn't be supporting model-server so, we plan to move that from the spec. 

> While I think it is ok, I want to know your perspective on having the trainer spec present in the exporter spec instead of being present in the Kepler spec alongside the exporter spec.

> The original idea is that if the Model server is ever deployed, the exporter will always be deployed but not necessarily the trainer.  The trainer will never be deployed without the exporter. Would there be situations in future where we only want the trainer deployed without the exporter?

The way I see it, Exporter, ModelServer, Trainer are different components, each having a set of features it exposes.  Keeping them separate allows configuring them separately. We shouldn't hard-code dependencies but (ab) using the child properties 

E.g foo requires bar should n't be represented as 
```
bar: 
  foo:
```


Rather keep them separate and have `foo` **refer** to `bar` using [`fieldRefs`](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/)  or other mechanisms 



```
foo:
  ... its properties
  bar:
    server: url

bar:
  ... bar's properties
```

PR #114, opened 2023-07-27
cleanup: support only exporter for the dev-preview

sthaha:
This addresses - https://github.com/sustainable-computing-io/kepler-operator/issues/115


husky-parul (approved):
lgtm

config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
```diff
@@ -37,55 +36,11 @@ spec:
...
-                        default: 3
-                        type: integer
-                      promSSLDisable:
-                        default: true
-                        type: boolean
-                    type: object
-                  url:
-                    default: ""
+                  version:
+                    default: latest
```
vprashar2929, line 43: I would suggest that we should use a stable version of Kepler as default since the latest might contain some developmental features

sthaha: Thats a todo before dev-preview is finalised

hack/cluster.sh
```diff
@@ -99,9 +99,11 @@ main() {
 
 	export CLUSTER_PROVIDER
 	export GRAFANA_ENABLE
+	cd "$DEV_CLUSTER_DIR"
 	"$DEV_CLUSTER_DIR/main.sh" "$op"
 
 	# NOTE: take additional actions after local-dev-cluster performs the "$OP"
+	cd "$PROJECT_ROOT"
 	on_cluster_"$op"
```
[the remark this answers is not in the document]
sthaha: I would rather keep that as a prerequisite to all functions that it can assume all file references to the root of the project.

hack/cluster.sh
```diff
@@ -99,9 +99,11 @@ main() {
 
 	export CLUSTER_PROVIDER
 	export GRAFANA_ENABLE
+	cd "$DEV_CLUSTER_DIR"
 	"$DEV_CLUSTER_DIR/main.sh" "$op"
```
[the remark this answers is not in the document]
sthaha: Actually, I want to fix the local-dev-cluster main.sh to cd to its project root so that external scripts can execute it without having to cd.

config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
sthaha: @vprashar2929 it would nice if you could drive the validation in `kepler` so that certain images can be tagged `stable` and then we can rely on `stable` tag here.

PR #117, opened 2023-07-27
Update bundle manifests

vprashar2929:


bundle/manifests/kepler-operator.clusterserviceversion.yaml
```diff
@@ -347,9 +333,11 @@ spec:
...
   - email: hchen@redhat.com
     name: Huamin Chen
+  - email: sthaha@redhat.com
+    name: Sunil Thaha
   maturity: alpha
   provider:
     name: Kepler Operator Contributors
     url: https://sustainable-computing.io/
-  version: 0.4.1
   replaces: kepler-operator.v0.4.0
```
husky-parul, line 342: `replaces: kepler-operator.v0.4.1`

husky-parul (changes requested):
**Note:** This csv file cannot be used as it is for the community operator. It needs to be further updated for the community operator PR because additional fields has to be added to conform to the community operator csv guidelines. For e.g. `icon` is one of those fields.

bundle/manifests/kepler-operator.clusterserviceversion.yaml
husky-parul: I did some more reading. We need to update the `replaces` because we use `replaces-mode` instead of `semver-mode` in the ci.yaml.  [relaces-mode](https://redhat-openshift-ecosystem.github.io/community-operators-prod/operator-ci-yaml/#replaces-mode) is the default so if we switch to `semver-mode` updating the `replaces` will not be required.

sthaha: Although this is true, I believe it is best to stick to replaces-mode until we know for use that with `semver` API breakages are okay or have a clear idea on how to specify it in the bundle manifests.

husky-parul: > `replaces: kepler-operator.v0.4.1` `replaces` is for replacing the previous csv version.

@vprashar2929 If you can fix the `version` and `replaces` property we can close this PR.
Thanks.

husky-parul (changes requested):
Tiny fix on `version` and `replaces`, everything else LGTM

husky-parul (approved):

PR #118, opened 2023-07-27
fix repo name

vprashar2929:
This PR fixes the repo name inside Makefile


husky-parul (approved):

sthaha (commented):
Thank you for this patch @vprashar2929 🙇

PR #119, opened 2023-07-28
feat: only watch the namespace where components are deployed

sthaha:


sthaha:
@husky-parul I have tested this on openshift 4.13 and works fine :)

husky-parul (approved):

PR #123, opened 2023-08-01
cleanup: remove unused RBAC from operator

sthaha:


config/rbac/role.yaml
```diff
@@ -21,14 +19,11 @@ rules:
...
   resources:
   - configmaps
-  - persistentvolumeclaims
-  - persistentvolumes
   - serviceaccounts
   - services
   verbs:
   - create
   - delete
-  - get
```
sthaha, line 31: @husky-parul  note that list and watch should be sufficient

husky-parul (approved):

Issue #87, opened 2023-07-10
Enhance e2e testing for kepler-operator


husky-parul:
Currently the e2e test is very limited. It only tests on kind and only test if the kepler instances are running. 
The e2e tests need to be updated to cover 
- [x] ~~cover other env like microshift~~
- [ ]  openshift
- [x]  to support ~microshift and~ openshift we might need to investigate installation through catalog
- [ ] #165

**Update**
In the discussion on streamlining kepler-operator deployment across multiple platform like OCP, kubernetes (kind) and ~microshift~ we came to the following conclusions:

- Operator installation should be done through Catalog.
- In OCP environment, the user can use the community catalog to install the operator
- In scenario where community catalog is not available (~microshift~air-gapped etc) users will deploy kepler by building local file catalog or adding the CatalogSource to their cluster. 

This will ensure consistent deployment across all platforms and ensure that all deployments are tested via the e2e tests in the CI.

Following issues will be also addressed:

- [x] #104 

This issue is relies on `operator-sdk` in local dev cluster to run olm suggested in this issue https://github.com/sustainable-computing-io/local-dev-cluster/issues/39

@sustainable-computing-io/keperl-ci-maintainer @sustainable-computing-io/maintainer 
@sthaha @vprashar2929

husky-parul:
@vprashar2929 
We can use this issue for e2e testing of the operator

husky-parul:
@vprashar2929 @sthaha 
What fw can be use in the upstream. Initially we started with `kuttl` but since it is limited to kind we didn't make much progress. If these tests are yaml based (both in the upstream and downstream) we can re use the artifacts in that case. Wdyt?

sthaha:
What we did in ObO was to just `go test` itself to write all e2e tests. 
 Writing tests are fairly straight forward and the test we write  isn't tied to any particular flavour of k8s. 

see: https://github.com/rhobs/observability-operator/blob/9b663b73963cf94a91b784680eb5b0cf4aa7c41a/test/e2e/monitoring_stack_controller_test.go#L214

husky-parul:
@vprashar2929 Can you please advise on how we can do tests for OpenShift using their CI? 
(If we can get openshift-ci  integrated with the sustainble-computinng org)

sthaha:
> cover other env like microshift

We are removing microshift since we intent to have multiple nodes available in our CI which is only supported by kind.

vprashar2929:
closing this as it is not relevant

Issue #104, opened 2023-07-24
Use bundle image in CI to deploy Operator


sthaha:
Since Operator is deployed in OpenShift using OLM, it is best to test if the operator deployment works as expected using 
`opeators-sdk run bundle` which deploys the operator and sets up the RBAC defined in CSV.

husky-parul:
@sthaha 
We couldn't incorporate this test because we didn't have an OCP CI environment. Is this going to be part of kepler-operator CI that runs on https://github.com/openshift/openshift-tests-private ?

sthaha:
The idea is to run `operator-sdk olm install` to setup OLM on kind and then use `make bundle bundle-image bundle-push` to push images to local registry, use  `operator-sdk run bundle`  to install operator using OLM in the cluster 
See for reference: https://github.com/rhobs/observability-operator/blob/main/.github/e2e-tests-olm/action.yaml#L38

husky-parul:
@sthaha 
I was looking into the e2e test. So the tests get called from 

    go test -v -failfast -timeout $TEST_TIMEOUT ./test/e2e/... --retain=true | tee "$LOGS_DIR/e2e.log" || ret=1

https://github.com/rhobs/observability-operator/blob/23a60b4789520c3c9087c1987a28fa1d637578b7/test/run-e2e.sh#L157

I am curious why we used go for setting up the environments?  Like creating all the resources and setting up the monitoring stack etc...

sthaha:
> I am curious why we used go for setting up the environments? Like creating all the resources and setting up the monitoring stack etc...

Using go to create the resource was only a shortcut since `oc` or `kubectl` would essentially do the same and we already had helpers to do that. I am okay with using a wrapper around `oc | kubectl`.  

Using Go for verification of the resources created is a big win since all assertions are made on static type.

PR #124, opened 2023-08-03
Setup e2e test framework

husky-parul:
This PR :
- adds olm build, test and push of the operator bundle image in a GHA 
- closes:
        https://github.com/sustainable-computing-io/kepler-operator/issues/87 partially
        https://github.com/sustainable-computing-io/kepler-operator/issues/104


sthaha:
@husky-parul is this good for review? If so, could you please rebase this onto #127 so that we can reuse `hack/utils.bash`

.github/workflows/e2e-olm/e2e-tests-olm.yml
```diff
@@ -0,0 +1,79 @@
...
+  pull_request:
+
+    branches:
+    - v1alpha1
+
+env:
+  KIND_VERSION: "0.15.0"
+  GO111MODULE: "on"
+  OPERATOR_IMG: "quay.io/sustainable_computing_io/kepler-operator"
+  KUBECONFIG: /tmp/.kube/config
```
sthaha, line 13: shouldn't this change depending on the `PROVIDER`? 
FYI local-dev-cluster scripts now copy <provider>/kubconfig to ~/.kube/config so this may not be even required.

.github/workflows/e2e-olm/e2e-tests-olm.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Install Go
+        uses: actions/setup-go@main
+        with: 
+          go-version-file: go.mod
+
+
+      - name: Use kepler action for kind cluster build
+        uses: sustainable-computing-io/kepler-action@v0.0.1
+        with:
+          runningBranch: ${{matrix.cluster_provider}}
```
sthaha, line 39: could you please help understand why `runningBranch` is needed?

.github/workflows/e2e-olm/e2e-tests-olm.yml
```diff
@@ -0,0 +1,79 @@
...
+          cluster_provider: ${{matrix.cluster_provider}}
+
+      - name: Install tools
+        run: |
+          chmod +x tests/run-e2e.sh
```
sthaha, line 44: we should already set `+x` and git add it, removing the need to do in CI.

.github/workflows/e2e-olm/e2e-tests-olm.yml
```diff
@@ -0,0 +1,79 @@
...
+          ./test/run-e2e.sh tools_install
```
sthaha, line 45: How about 
```suggestion
          ./test/run-e2e.sh install-tools 
```
Also, can't the script check for the tools required and auto install those?

.github/workflows/e2e-olm/e2e-tests-olm.yml
vprashar2929: [runningBranch](https://github.com/sustainable-computing-io/kepler-action/blob/d7e47ee387e8d26c4d7bf0b2232c90f845dd80e2/index.js#L111) was used in kepler-action to perform a particular function. I think after [this](https://github.com/sustainable-computing-io/kepler-action/pull/63) we don't need to provide runningBranch in our workflows.

PR #126, opened 2023-08-06
Update cluster.sh to use operator-sdk from make tools

vprashar2929:
This PR updates the `cluster.sh` to use `operator-sdk` binary which is installed as part of `make tools`. This allows the script to use the specific version of `operator-sdk` hence not interfering with the version installed on the user's machine

### DIFF
--- a/hack/cluster.sh
+++ b/hack/cluster.sh
@@ -29,6 +29,7 @@ PROJECT_ROOT="$(git rev-parse --show-toplevel)"
 declare -r PROJECT_ROOT
 declare -r TMP_DIR="$PROJECT_ROOT/tmp"
 declare -r DEV_CLUSTER_DIR="$TMP_DIR/local-dev-cluster"
+declare -r BIN_DIR="$TMP_DIR/bin"
 
 info() {
 	echo -e " 🔔 $*" >&2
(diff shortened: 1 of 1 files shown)


hack/cluster.sh
```diff
@@ -75,7 +76,12 @@ on_cluster_up() {
 	kubectl apply --force -f "$PROJECT_ROOT/hack/crds"
 
 	info "setup OLM"
-	operator-sdk olm install --verbose
+	[[ -f "$BIN_DIR/operator-sdk" ]] || {
+		err "operator-sdk not available"
+		info "run make tools"
+		return 1
+	}
```
sthaha, line 83: +1 Nice addition! 

but I don't think we should insist that operator-sdk is always installed to `"$BIN_DIR"` it is possible that user has installed elsewhere. We should 
1. validate if the binary exists and if the version is the right one. see makefile. Or just call `make operator-sdk` which already does this.
2. in main(), `export PATH="$BIN_DIR:$PATH"`

vprashar2929: > I don't think we should insist that operator-sdk is always installed to "$BIN_DIR" it is possible that user has installed elsewhere

we already recommend running `make tools` before performing any operations as it allows us to set up the required tools with the versions we want on the user's local machine. And it assumes installing those tools inside `BIN_DIR`. Why should we consider if operator-sdk is available or not and what version of operator-sdk is pre-installed on the user's machine?

sthaha: >  Why should we consider if operator-sdk is available or not and what version of operator-sdk is pre-installed on the user's machine?

If the operator-sdk is already installed, and matches the version, what is the in point of insisting it be installed in a specific path? This become even more of an issue if you use git work-trees.

sthaha (approved):
Thanks a lot @vprashar2929 🙇

Issue #122, opened 2023-07-31
Fix grafana dashboard setup manifests


vprashar2929:
Post-refactor changes the current manifests which are present inside `hack/dashboard/openshift/` don't work correctly. The manifests should reflect the refactor changes such as namespace changes, selector label changes, etc.

PR #127, opened 2023-08-07
feat: improve grafana deployment

sthaha:
The commit improves Grafana deployment on OpenShift by
- Adapting to the changes introduced by the refactor commmit
- Making backup of CMO config for later restoration
- Providing key information about the deployment

Additionally, common bash utilities have been moved to hack/utils.bash

Fixes: https://github.com/sustainable-computing-io/kepler-operator/issues/122


sthaha:
@vprashar2929 Could you please help verify if this works for you? 
cc: @husky-parul

sthaha:
cc: @vimalk78

hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -1,56 +1,240 @@
...
+}
+
+setup_grafana() {
+	header "Setting up Grafana"
+
+	info "Creating Grafana resources"
+	oc_apply_kepler_ns -k hack/dashboard/openshift/grafana-deploy
+
+	wait_until 20 10 "Grafana to be deployed" \
+		oc get grafana,grafanadatasource
```
vprashar2929, line 146: oc get should be in the namespace where grafana and grafanadatasource are deployed. Currently, it checks in the default namespace

hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -1,56 +1,240 @@
...
+	}
+
+	ok "Backed up $CMO_CM config to $BACKUP_DIR"
+}
+
+setup_grafana() {
+	header "Setting up Grafana"
+
+	info "Creating Grafana resources"
+	oc_apply_kepler_ns -k hack/dashboard/openshift/grafana-deploy
```
vprashar2929, line 143: Is there any specific reason to deploy grafana inside openshift-kepler-operator ns? Shouldn't we isolate this?

hack/utils.bash
```diff
@@ -0,0 +1,105 @@
...
+wait_until() {
+	local max_tries="$1"
+	local delay="$2"
+	local msg="$3"
+	local condition="$4"
+	shift 4
+
+	info "Waiting [$max_tries x ${delay}s] for $msg"
+	local tries=0
+	local max_tries=20
```
vprashar2929, line 90: Why we are assigning `max_tries=20` here?

hack/utils.bash
```diff
@@ -0,0 +1,105 @@
...
+# wait_until <max_tries> <delay> <msg> <condition>
+# waits for condition to be true for a max of <max_tries> x <delay> seconds
+wait_until() {
+	local max_tries="$1"
+	local delay="$2"
+	local msg="$3"
+	local condition="$4"
+	shift 4
+
+	info "Waiting [$max_tries x ${delay}s] for $msg"
```
vprashar2929, line 88: Can we just specify `Waiting for "$msg"`. We are anyway showing max_tries and delay at line 100?

hack/utils.bash
sthaha: good catch!!

hack/utils.bash
sthaha: I feel the initial comment is useful in understanding how much it would wait.

hack/dashboard/openshift/deploy-grafana.sh
sthaha: it is namespace agnostic so shouldn't matter which ns it is, so it is better to no specify one.

hack/dashboard/openshift/deploy-grafana.sh
sthaha: I don't have a strong opinion on this. The intention is that when `kepler` is removed, the ns is removed and thus grafana is removed as well.

hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -1,56 +1,240 @@
...
+	line 55 heavy
+}
+
+main() {
+	cd "$PROJECT_ROOT"
+	mkdir -p "$BACKUP_DIR"
+
+	validate_cluster || {
+		fail "Cluster validation failed!"
+		info" Fix issues reported above and rerun the script"
```
vprashar2929, line 227: should be space after info.  like `info "Fix issues reported above and rerun the script"`

hack/utils.bash
```diff
@@ -0,0 +1,104 @@
...
+			ret=0
+			break
+		}
+
+		tries=$((tries + 1))
+		echo "   ... [$tries / $max_tries] waiting ($delay secs) - $msg" >&2
+		sleep "$delay"
+	done
+
+	return $ret
```
vprashar2929, line 103: Shouldn't we make use of ret where wait_until is called?

sthaha: Like where exactly and why :) ?

vimalk78 (approved):
/lgtm

PR #129, opened 2023-08-08
fix make bundle

vprashar2929:
This PR:
1. adds support to make yq part of make tools.
2. change the directory from config/manager to config/manifests while setting the image controller.
3. move the bundle.Dockerfile to bundles dir post generating bundle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,9 +151,9 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: deploy
 deploy: install ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && \
-		$(KUSTOMIZE) edit set image controller=$(OPERATOR_IMG)
-	$(KUSTOMIZE) build config/default | kubectl apply --server-side --force-conflicts -f -
+	$(KUSTOMIZE) build config/default | \
+		sed  -e "s|<OPERATOR_IMG>|$(OPERATOR_IMG)|g" | tee tmp/deploy.yaml | \
+		kubectl apply --server-side --force-conflicts -f -
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -269,16 +286,17 @@ endif
 .PHONY: bundle
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
 	$(OPERATOR_SDK) generate kustomize manifests --apis-dir=./pkg/api --verbose
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(OPERATOR_IMG)
+	cd config/manifests && $(KUSTOMIZE) edit set image controller=$(OPERATOR_IMG)
 	$(KUSTOMIZE) build config/manifests | tee tmp/pre-bundle.yaml  |  \
 		$(OPERATOR_SDK) generate bundle --version $(VERSION) $(BUNDLE_GEN_FLAGS)
 	sed < bundle/manifests/kepler-operator.clusterserviceversion.yaml "s|containerImage.*|containerImage: $(OPERATOR_IMG)|g" \
```
sthaha, line 292: is this needed? doesn't the `set image controller` above not work?

config/manifests/kustomization.yaml
```diff
@@ -1,27 +1,15 @@
...
-#    namespace: system
-#  patch: |-
-#    # Remove the manager container's "cert" volumeMount, since OLM will create and mount a set of certs.
-#    # Update the indices in this path if adding or removing containers/volumeMounts in the manager's Deployment.
-#    - op: remove
-#      path: /spec/template/spec/containers/1/volumeMounts/0
-#    # Remove the "cert" volume, since OLM will create and mount a set of certs.
-#    # Update the indices in this path if adding or removing volumes in the manager's Deployment.
-#    - op: remove
-#      path: /spec/template/spec/volumes/0
```
sthaha, line 27: Could you please maintain this? We will need webhooks soon to disallow creation of kepler instances other than kepler.

sthaha (changes requested):
This commit does 3 things. We should break that in to 3 commits. 
1. patch that installs yq
2. fix the bundle
3. remove webhook part that is commented. 

Why is this important? 
lets say we want to revert the `fix bundle part` or `remove webhook` part, `git revert <commit>` is all that is needed. You can make 3 different commits from the same file using `git add -p` to select parts that form part of the commit. 


Do note the main reason for holding this PR is to make sure that we need / do not need the `sed`

vprashar2929 (commented):
LGTM

sthaha (approved):

PR #132, opened 2023-08-09
fix(grafana): validate version of oc

sthaha:


hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -28,6 +28,24 @@ validate_cluster() {
...
+	oc_minor_version=$(echo "$oc_version" | cut -f2 -d.)
+
+	[[ $oc_major_version -eq 4 ]] && [[ $oc_minor_version -ge 12 ]] || {
+		fail "oc version $oc_version used should be at least 4.12.0"
+		info "install a newer version of oc"
+		cat <<-EOF
+			     curl -sNL https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.13.0/openshift-client-linux.tar.gz |
+				   tar -xzf - -C .
+		EOF
+		ret=1
```
vimalk78, line 46: will the script continue from here onwards?

hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -28,6 +28,24 @@ validate_cluster() {
...
+	local -i oc_major_version oc_minor_version
+	oc_major_version=$(echo "$oc_version" | cut -f1 -d.)
+	oc_minor_version=$(echo "$oc_version" | cut -f2 -d.)
+
+	[[ $oc_major_version -eq 4 ]] && [[ $oc_minor_version -ge 12 ]] || {
+		fail "oc version $oc_version used should be at least 4.12.0"
+		info "install a newer version of oc"
+		cat <<-EOF
+			     curl -sNL https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.13.0/openshift-client-linux.tar.gz |
+				   tar -xzf - -C .
```
vprashar2929, line 44: can we do this `tar -xzf - -C /usr/local/bin` ?

hack/dashboard/openshift/deploy-grafana.sh
sthaha: yes, but the return value will be 1 so that script function will continue with all validations and then exit
```
   🔆🔆🔆  Validating cluster  🔆🔆🔆
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Error from server (NotFound): namespaces "openshift-monitoring" not found
 ❌ FAIL: openshift-monitoring namespace missing. Is this an OpenShift cluster?
 🔔 cluster:  https://127.0.0.1:40853
 ❌ FAIL: Missing Kepler CRD. Is Kepler Operator Installed?
error: the server doesn't have a resource type "kepler"
 ❌ FAIL: Mising kepler deployment. Did you create kepler instance?
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ❌ FAIL: Cluster validation failed!
 🔔 Fix issues reported above and rerun the script
```

hack/dashboard/openshift/deploy-grafana.sh
sthaha: I wouldn't encourage putting anything in system path. It is best for users to decide where the binary should be installed.

Again, this isn't the only way to install oc.

vprashar2929 (approved):
LGTM

PR #139, opened 2023-08-14
run e2e test only against kind

sthaha:


hack/cluster.sh
```diff
@@ -90,14 +96,24 @@ main() {
 	local op="$1"
 	shift
 	cd "$PROJECT_ROOT"
-
-	# NOTE: all operations are relative to tmp
+	export PATH="$BIN_DIR:$PATH"
 	mkdir -p "${TMP_DIR}"
-	git_checkout
+	ensure_all_tools
```
vprashar2929, line 101: Do we need to call ensure_all_tools at every operation? I think we should call this during cluster_prereqs step or during on_cluster_up. Wdyt?

hack/cluster.sh
```diff
@@ -90,14 +96,24 @@ main() {
...
+
+	#NOTE: allow cluster_<OP> to executed without going through the entire
+	# setup again. This is useful (especially) in CI which uses the kepler-action
+	# to setup a cluster which does not have all the prerequisites for running
+	# Operator as OLM Bundle
+	declare -F | cut -f3 -d ' ' | grep -qE "^cluster_$op\$" && {
+		header "Running cluster_$op"
+		"cluster_$op" "$@"
```
vprashar2929, line 109: Is there any need to pass args to cluster_prereqs?

tests/run-e2e.sh
```diff
@@ -18,8 +25,155 @@ declare SHOW_USAGE=false
...
+			grep -i error | tee "$error_log"
+		line 50
+	fi
+
+	return $ret
+}
+
+parse_args() {
+	### while there are args parse them
+	while [[ -n "${1+xxx}" ]]; do
```
vprashar2929, line 140: In case of no args passed while running `run-e2e.sh` shouldn't we show help and exit?

hack/cluster.sh
sthaha: this applies to all `cluster_op` so it is best to pass all args to that function .

tests/run-e2e.sh
sthaha: no, `./test/run-e2e.sh` is a valid invocation.

hack/cluster.sh
sthaha: yes, always best to ensure all tools are present. This is an assumption that all functions can make. thus no need to check if `x, y, z` is installed.

vprashar2929: Why need this in case of cluster_down?

sthaha: The same applies,  ... it is best to ensure all tools exist before starting anything and for a non - production script, these things do not matter.

hack/cluster.sh
vprashar2929: We only have cluster_prereqs and we don't use args there. What's the point of passing args there?

vprashar2929:
Also please make change [here](https://github.com/sustainable-computing-io/kepler-operator/blob/c6ee1f49e837221c506bb2e649f155a2c9fcc6b6/tests/run-e2e.sh#L192) as well😅

tests/run-e2e.sh
```diff
@@ -18,8 +25,155 @@ declare SHOW_USAGE=false
...
+	kubectl get events -w \
+		-o custom-columns=FirstSeen:.firstTimestamp,LastSeen:.lastTimestamp,Count:.count,From:.source.component,Type:.type,Reason:.reason,Message:.message \
+		-n "$ns" | tee "$LOGS_DIR/$ns-events.log"
+}
+
+watch_operator_errors() {
+	local err_log="$1"
+	shift
+
+	kubectl logs -f -n "$OPERATORS_NS" "deploy/$OPERATOR_DEPLOY_NAME" |
```
vprashar2929, line 102: We shouldn't use -f here.

vprashar2929:
We can fix some of the issues within the script in later PR but the initial implementation works fine

vprashar2929 (approved):
LGTM

tests/run-e2e.sh
sthaha: we should use -f here :) .. it looks only for errors. See the invocation below where this is run in background.

tests/run-e2e.sh
sthaha: When you run locally which does a complete build and deploy of the operator and bundle.

Issue #137, opened 2023-08-11
remove spec.exporter.Version field from Kepler spec


sthaha:
Currently user is allowed to specify any version of Kepler and operator deploys the image. Such a feature is unmaintainable until Kepler become GA. (1.0.0) and doesn't introduce breaking changes.
A pragmatic approach will be to pin the version of the kepler being deployed 
#### Things to consider 
* argument to operator `--image-kepler=<img-repo-url>`
* A confimap specifically for operator to store operator specific information 
* May be a config CRD for operator (since we have time and again seen that configmap soon turn into a CRD) 
*

PR #146, opened 2023-08-17
feat!: pin kepler version to 0.5.4

sthaha:
Fixes: https://github.com/sustainable-computing-io/kepler-operator/issues/137

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -35,6 +35,7 @@ import (
 
 	keplersystemv1alpha1 "github.com/sustainable.computing.io/kepler-operator/pkg/api/v1alpha1"
 	"github.com/sustainable.computing.io/kepler-operator/pkg/components"
+	"github.com/sustainable.computing.io/kepler-operator/pkg/components/exporter"
 	"github.com/sustainable.computing.io/kepler-operator/pkg/controllers"
 	//+kubebuilder:scaffold:imports
 )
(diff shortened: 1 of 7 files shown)


cmd/manager/main.go
```diff
@@ -53,15 +54,29 @@ func init() {
...
+
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+
+	// NOTE: KEPLER_IMAGE can be set as env or flag, flag takes precedence over env
+	keplerImage := env("KEPLER_IMAGE", exporter.StableImage)
+	flag.StringVar(&exporter.Config.Image, "kepler.image", keplerImage, "kepler image")
```
sthaha, line 78: perhaps we should extract this pattern out if we add more options. The idea being that something can be both env var or a flag.

husky-parul (approved):

pkg/components/exporter/exporter.go
```diff
@@ -48,8 +48,16 @@ const (
 	DaemonSetName = prefix + "ds"
 	ServiceName   = prefix + "svc"
 
-	// TODO: make this config to the operator
-	imageRepo = "quay.io/sustainable_computing_io/kepler:"
+	StableImage = "quay.io/sustainable_computing_io/kepler:release-0.5.4"
```
vprashar2929, line 51: Does this mean we will go with `release-0.5.4` for dev-preview ?

sthaha: It depends on the outcome of the testing :)

PR #155, opened 2023-08-21
fix: use correct data-type of exporter port - int32

sthaha:


sthaha:
cc @vimalk78

pkg/components/exporter/exporter.go
```diff
@@ -172,7 +172,7 @@ func NewConfigMap(d components.Detail, k *v1alpha1.Kepler) *corev1.ConfigMap {
 	}
 
 	exporter := k.Spec.Exporter
-	bindAddress := "0.0.0.0:" + strconv.Itoa(exporter.Port)
+	bindAddress := "0.0.0.0:" + strconv.Itoa(int(exporter.Port))
```
vimalk78, line 175: same as above

pkg/components/exporter/exporter.go
```diff
@@ -74,7 +74,7 @@ var (
 func NewDaemonSet(k *v1alpha1.Kepler) *appsv1.DaemonSet {
 	exporter := k.Spec.Exporter
 
-	bindAddress := "0.0.0.0:" + strconv.Itoa(exporter.Port)
+	bindAddress := "0.0.0.0:" + strconv.Itoa(int(exporter.Port))
```
vimalk78, line 77: type cast to `int` or `int32` ?

sthaha: Itoa requires int hence the conversion.

vimalk78 (approved):
looks good to me

PR #159, opened 2023-08-21
fix: add node selector to select linux nodes

vimalk78:
tested on openshift 4.13


vprashar2929 (approved):
LGTM

pkg/components/exporter/exporter.go
```diff
@@ -69,6 +69,10 @@ var (
 	podSelector = labels.Merge(k8s.StringMap{
 		"app.kubernetes.io/name": "kepler-exporter",
 	})
+
+	LinuxNodeSelector = k8s.StringMap{
```
sthaha, line 73: we shouldn't export variables unless there is a good reason

PR #176, opened 2023-08-25
fix: validate exporter.port range

sthaha:


pkg/api/v1alpha1/kepler_types.go
```diff
@@ -95,6 +95,8 @@ type ExporterSpec struct {
 	// TODO: fix the default version before dev-preview
 
 	// +kubebuilder:default=9103
+	// +kubebuilder:validation:Maximum=65535
+	// +kubebuilder:validation:Minimum=1
```
vimalk78, line 99: a user created port cannot take values as low as 1

sthaha:
```
 ▲ kop/wt/e2e-test oc apply -f tmp/invalid.yaml                                                        ⇡ e2e-test 7m ● ⬡
The Kepler "invalid-name" is invalid: spec.exporter.port: Invalid value: 910322222: spec.exporter.port in body should be less than or equal to 65535
 △ kop/wt/e2e-test                                                                                                                                                                    
```

pkg/api/v1alpha1/kepler_types.go
sthaha: why not? should it be > 1024?

sthaha: I have validated, and it works, could you please check if that works for you? Try setting the port to 1 and

```
 kop/wt/e2e-test kubectl port-forward services/kepler-exporter-svc 3000:1 --address=0.0.0.0 -n openshift-kepler-operator
```
Forwarding from 0.0.0.0:3000 -> 1

```
curl http://localhost:3000/metrics | ag 'kepler_'
```

sthaha: Whatever may be case be, this should be reflected in the `Availablility` status. For the record, I am open to increasing the min value. But do note that for k8s service, the allowed range starts from `1`

vimalk78: i think so, creating a lower valued port needs SELinux permissions

vprashar2929 (approved):
LGTM

PR #180, opened 2023-08-27
ci: validate images can be built with right version pre-merge

sthaha:
This patch validates that the images can be built and that the bundle
is validated by
  * refactoring version computation
  * refactoring image builds to a separate action that is reused

### DIFF
--- a/.github/compute-version/action.yaml
+++ b/.github/compute-version/action.yaml
@@ -1,0 +1,16 @@
+name: Compute Version
+description: computes version of the operator
+outputs:
+  version:
+    description: "operator version"
+    value: ${{ steps.version.outputs.version }}
+runs:
+  using: composite
+  steps:
+    - name: Generate Version
+      id: version
+      shell: bash
+      run: |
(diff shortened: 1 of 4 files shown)


sthaha:
@vprashar2929 , could you please review ?

sthaha:
@vprashar2929 here is the run in my fork (post merge) https://github.com/sthaha/kepler-operator/actions/runs/5995816722

.github/compute-version/action.yaml
```diff
@@ -0,0 +1,16 @@
+name: Build and Publish Images
+description: 'Publishes operator and bundle images to an Image Registry'
```
vprashar2929, line 2: Can we change the name and description of this workflow to align with computing the version?

sthaha: Good catch :)

sthaha:
@vprashar2929 could you please take another look?

vprashar2929 (approved):
LGTM

Issue #130, opened 2023-08-08
Make Kepler deployed on nodes with suitable label attached


vprashar2929:
Currently, when a Kepler instance is created from Operator it by default deploys Kepler-exporter on all the available nodes on the cluster. Since there is an older[ pre-req](https://github.com/sustainable-computing-io/kepler/tree/main/manifests#installing-kepler-on-openshift) which allows flexibility to deploy Kepler on either all nodes or some specific nodes which user wants. 

This enhancement can be added in Operator to look for labels that are attached to nodes and on the basis of that deploy Kepler on respective nodes.

sthaha:
@vprashar2929 , I think the better idea would be to allow `nodeSelector` and `tolerations` to be added to the `kepler` daemonset. E.g.

```
kind: Kepler
spec:
  exporter: 
     nodeSelector: {...}
     tolerations: {...}
```

PR #181, opened 2023-08-28
feat: add Node Selector and Tolerations

vimalk78:
Fixes: https://github.com/sustainable-computing-io/kepler-operator/issues/130
Fixes POWERMON-118


pkg/api/v1alpha1/kepler_types.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	v1 "k8s.io/api/core/v1"
```
sthaha, line 20:
```suggestion
	corev1 "k8s.io/api/core/v1"
```
To follow the convention

pkg/components/exporter/exporter.go
```diff
@@ -99,17 +107,10 @@ func NewDaemonSet(k *v1alpha1.Kepler) *appsv1.DaemonSet {
...
 					DNSPolicy:          corev1.DNSPolicy(corev1.DNSClusterFirstWithHostNet),
-					Tolerations: []corev1.Toleration{
-						{
-							Effect: corev1.TaintEffectNoSchedule, Key: "node-role.kubernetes.io/master",
-						},
-						{
-							Effect: corev1.TaintEffectNoSchedule, Key: "node-role.kubernetes.io/infra",
-						},
-					},
+					Tolerations:        append(defaultTolerations, tolerations...),
```
sthaha, line 113: I am not really sure of  ... I think we should apply the  toleration **only** if the user has not specified a toleration.
What if I want my deployment to run on normal nodes , i.e. do not run on master (for whatever reason).

pkg/components/exporter/exporter_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func TolerationsFromDS(ds *appsv1.DaemonSet) []corev1.Toleration {
+	return ds.Spec.Template.Spec.Tolerations
+}
+
+func TestNodeSelection(t *testing.T) {
+
+	tt := []struct {
+		spec     ko.ExporterSpec
```
sthaha, line 23: could you please add a scenario? 
```suggestion
		scenario   string
		spec     ko.ExporterSpec
```
And use 
```
t.Run(tc.scenario, func(...){
   t.Parallel()

})
```

pkg/components/exporter/exporter_test.go
```diff
@@ -0,0 +1,75 @@
+package exporter
+
+import (
+	"testing"
+
+	ko "github.com/sustainable.computing.io/kepler-operator/pkg/api/v1alpha1"
```
sthaha, line 6:
```suggestion
	"github.com/sustainable.computing.io/kepler-operator/pkg/api/v1alpha1"
```
To match the convention 👼

pkg/components/exporter/exporter_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"gotest.tools/v3/assert"
```
sthaha, line 7: nit: we should rely on just one assert package. I am okay with any :) but pick one and use that.

vimalk78: is there something already chosen? i will use that.

pkg/components/exporter/exporter.go
vimalk78: users can add another taint to exclude master nodes in that case

pkg/api/v1alpha1/kepler_types.go
vimalk78: ok

sthaha (changes requested):
Looks good overall, few minor comments to address.

pkg/components/exporter/exporter_test.go
vimalk78: ok

pkg/api/v1alpha1/kepler_types.go
```diff
@@ -98,6 +99,14 @@ type ExporterSpec struct {
...
 	// +kubebuilder:validation:Minimum=1
 	Port int32 `json:"port,omitempty"`
+
+	// Defines which Nodes the Pod is scheduled on
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// If specified, define Pod's tolerations
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```
sthaha, line 109: Just wondering if we can set the defaults here?

vimalk78: in code it is fine too, not too much difference.

sthaha (approved):
Looks great! 
There is a suggestion about setting the defaults for `nodeSelector` and `tolerations` but we can follow this up in a different PR.

PR #188, opened 2023-08-30
remove e2e-success step from workflow

vprashar2929:
This PR removes the redundant e2e-success step from [integration-test](https://github.com/sustainable-computing-io/kepler-operator/blob/31e6dff51eee3c6a1093b470747c0d1c561a5597/.github/workflows/integration_test.yml#L46) workflow


sthaha (approved):
Thank you!

PR #190, opened 2023-08-31
Improve default target

sthaha:
Default make target (all) now brings up a cluster, builds operator and
bundle images, provides instructions on how to run the operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests generate fmt vet  ## Run tests.
+test:  fmt vet  ## Run tests.
 		go test ./pkg/... -coverprofile cover.out
 
 ##@ Development env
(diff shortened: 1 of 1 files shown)


sthaha:
@vprashar2929 🙏

Makefile
```diff
@@ -3,6 +3,8 @@
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+THIS_FILE := $(lastword $(MAKEFILE_LIST))
```
vprashar2929, line 6: Didn't know we could include make files like this😅

Makefile
```diff
@@ -35,7 +37,23 @@ IMG_BASE ?= quay.io/sustainable_computing_io
 OPERATOR_IMG ?= $(IMG_BASE)/kepler-operator:$(VERSION)
 
 .PHONY: all
-all: build
+all: ## default target - setup a k8s cluster with operator installed
```
vprashar2929, line 40: cluster-up doesn't take into account the env variable that we set [here](https://github.com/sustainable-computing-io/kepler-operator/blob/464c5221d96bc765833371b8e7510429889791bf/Makefile#L80). It brings up the cluster with default values.

vprashar2929:
Also, I was thinking shouldn't we exclude cluster-up from make all and let the user run independently according to their needs? Let make all be specific to building and pushing operator and bundle images at one go

Makefile
sthaha: is the default value fine? Note, this is for a "default" make target - all

vprashar2929 (approved):
LGTM

PR #191, opened 2023-08-31
feat: mount redfish credentials

vimalk78:
This PR mounts redfish credentials to the kepler exporter pod

### DIFF
--- a/pkg/components/exporter/exporter.go
+++ b/pkg/components/exporter/exporter.go
@@ -49,6 +49,8 @@ const (
 	ServiceName   = prefix + "svc"
 
 	StableImage = "quay.io/sustainable_computing_io/kepler:release-0.5.4"
+
+	RedfishSecretName = "redfish"
 )
 
 // Config that will be set from outside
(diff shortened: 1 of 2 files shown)


pkg/components/exporter/exporter.go
```diff
@@ -49,6 +49,8 @@ const (
 	ServiceName   = prefix + "svc"
 
 	StableImage = "quay.io/sustainable_computing_io/kepler:release-0.5.4"
+
+	RedfishSecretName = "redfish"
```
sthaha, line 53: Where are we creating the secret ? 🤔

pkg/components/exporter/exporter.go
```diff
@@ -175,6 +178,7 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.Kepler) *appsv1.DaemonSe
 						k8s.VolumeFromHost("proc", "/proc"),
 						k8s.VolumeFromHost("kernel-debug", "/sys/kernel/debug"),
 						k8s.VolumeFromHost("kernel-src", "/usr/src/kernels"),
+						k8s.VolumeFromSecret("redfish-cred", RedfishSecretName),
```
sthaha, line 181: On a related notes, there is a down side to this that  we should solve .. which is that secrets will be deleted when `kepler` is deleted :( ... Need to find a way to optionally delete ns if user created objects (including grafana exists).

pkg/components/exporter/exporter.go
sthaha: If the redfish secret is optional, how about we allow 
```
exporter:
  redfish:
    secretRef: 
```

Also, why are we using secret instead of say configmap?

sthaha (changes requested):
Lets discuss this further before providing this feature.

sthaha:
closing this since #364 is merged 🎊

PR #212, opened 2023-09-07
feat: improve grafana UX

sthaha:
This patch improves the grafana deployment experience by
* ensuring that kepler is available and print details if it isn't
* deploying grafana to `kepler-grafana` namespace so that deletion of kepler will not affect and grafana changes.
* Creating separate backup directory per run


hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -63,25 +74,53 @@ validate_cluster() {
...
+	oc get crds keplers.kepler.system.sustainable.computing.io || {
+		fail "Missing Kepler CRD. Is Kepler Operator Installed?\n"
 		ret=1
 	}
 
-	[[ $(oc get kepler kepler -o=jsonpath='{.status.conditions[?(@.type=="Available")].status}') == "True" ]] || {
-		fail "Mising kepler deployment. Did you create kepler instance?"
+	oc get kepler kepler >/dev/null 2>&1 || {
+		fail "Missing kepler instance. Did you forget to create kepler?"
+		info_run "oc apply -f ./config/samples/kepler.system_v1alpha1_kepler.yaml"
```
vprashar2929, line 90: why do we need to info this? Our sample yaml might keep on updating as and when new features are added

sthaha: Its only printing an information -right?

vprashar2929: agree but it might contain some configs that are experimental and not supported by the current operator version which is deployed. we can simply print that `create kepler instance from openshift console or follow our doc's`

vprashar2929 (approved):
LGTM

Issue #172, opened 2023-08-24
Kepler Operator Test Plan


vprashar2929:
# Installation
## Operator
**The scenarios related to Installation only cover the targeted version of OpenShift i.e 4.13**
### Positive:
- [ ] Kepler operator installs fine from Community Catalog
    > 👉 This is to be verified from OpenShift Console

- [ ] Kepler operator installs fine from Operator-SDK.
   > 👉 This is to be verified by Upstream/Downstream CI.

- [ ] Kepler operator should be able to deploy/handle any dependencies required.
    > 👉 This includes the ability of the operator to provision, configure, and manage resources such as pods, services, volumes, etc.

- [ ] Following resources should get deployed as part of Operator Installation:
    - [ ] `kepler-operator-controller-manager deployment`
        - [ ] `kube-rbac-proxy`
        - [ ] `manager`
    - [ ] `kepler-operator-controller-manager replicaset`
    - [ ] `kepler-operator-controller-manager-metrics-service`
    - [ ] `Kepler CRD`
    - [ ] `kepler-operator-controller-manager Service Account`
    - [ ] `roles`
    - [ ] `rolebindings`
    - [ ] `lease`
    - [ ] `servicemonitor` -> TBD
    - [ ] `kepler-operator CSV`
    - [ ] `installplan`
    - [ ] `subscription`
    - [ ] `packagemanifests`
    - [ ] `catalog`
### Negative:
- [ ] OpenShift console should log appropriate errors in case of unsuccessful installation of Operator.
  > 👉  This is to be verified from OpenShift Console.
           It can be related to any failure scenario or in case of unsuccessful installation.

### Positive: 
- [ ] Kepler operator when installed from Community Catalog can create a Kepler instance successfully.
  > 👉 This is to be verified from OpenShift console
          Kepler operator can reconcile `kepler` CR

- [ ] Kepler operator when installed from Operator-SDK can deploy Kepler instance when Kepler template is applied.
  > 👉 This is to be verified by Upstream/Downstream CI.

- [ ] Following resources should get deployed as part of Kepler Instance Creation:
    - [ ] `openshift-kepler-operator namespace`
    - [ ] `clusterrole`
    - [ ] `scc`
    - [ ] `clusterrolebindings`
    - [ ] `kepler-exporter-ds`
    - [ ] `kepler-exporter-cm`
    - [ ] `kepler-exporter-svc`
    
[...]

PR #216, opened 2023-09-07
doc: Add operator test-plan to docs

vprashar2929:
This PR moves test-plan from #172 to `docs/developer`


sthaha (approved):
Thank you for submitting this. We should require every feature added/removed to maintain this docs as well.

PR #217, opened 2023-09-07
feat: faster deletion reconciliation loop

sthaha:
Since deleting a namespace deletes all objects inside it, the reconciler only needs to delete the components namespace and wait for the deletion to complete.


vimalk78 (approved):
lgtm

PR #218, opened 2023-09-08
fix: removed old topk(10) namespace panel

vimalk78:
This PR removes the old `topk(10)` namespaces panel which was left for review in earlier PR #208 

new dashboard starts with
image


vprashar2929 (approved):
LGTM

PR #219, opened 2023-09-08
DNM: fix: change increase() to irate() in topk(10) namespace panel

vimalk78:
the `topk(10)` namespace panel uses `increase()` function. so the query gives increase in the joules over a period of 24 hours, and not watts as expected.
changed the `increase()` function to `irate()`  to get watts and then convert watts to kwh

kwh values with `increase()`:
image

kwh values with `irate()`:
image


vprashar2929 (approved):
LGTM

sthaha:
@vimalk78 , closing this since the queries have been fixed.

PR #221, opened 2023-09-11
feat: add git sha8 and branch name to image tags

sthaha:
This allows bundle to be traced back to the git sha and also allows 
`operator-sdk run bundle <bundle-image>:v1alpha1` to run latest v1alpha1 bundle


Makefile
```diff
@@ -35,6 +35,7 @@ IMG_BASE ?= quay.io/sustainable_computing_io
 # OPERATOR_IMG define the image:tag used for the operator
 # You can use it as an arg. (E.g make operator-build OPERATOR_IMG=<some-registry>:<version>)
 OPERATOR_IMG ?= $(IMG_BASE)/kepler-operator:$(VERSION)
+ADDITIONAL_TAGS ?=
```
vimalk78, line 38: during local development, does `operator-build`, `operator-push` get additional tags?

sthaha: since `ADDITIONAL_TAGS` is empty, it should not create any additional tags. Does this break on your local dev env?

vimalk78: did not test it yet.

vimalk78 (approved):
lgtm

PR #223, opened 2023-09-12
feat: operator deploys servicemonitor

sthaha:
This PR adds a feature to the operator to deploy a servicemonitor to monitor the kepler exporter service. 
It also modifies grafana deployment script to remove its deployment of servicemonitor and rely on the 
ServiceMonitor deployed by the operator.


sthaha:
@vprashar2929 FYI

go.mod
```diff
@@ -5,12 +5,13 @@ go 1.20
 require (
 	github.com/go-logr/logr v1.2.4
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.66.0
```
vimalk78, line 8: should we use a more recent version?

vimalk78 (approved):
lgtm

go.mod
sthaha: It is safer to follow what CMO has https://github.com/openshift/cluster-monitoring-operator/blob/master/go.mod#L16

sthaha: I did look into any API changes to ServiceMonitor from 0.66, didn't find any... however, didn't want to risk it either :D

BGrasnick:
This is such a welcome addition so we don't have to set it up and deploy it separately! I was just wondering if there are any uses cases / scenarios where users might not want or need the ServiceMonitor? If so would it be possible to make it configurable but deploy by default? Just some thoughts :)

sthaha:
@BGrasnick  thanks, yes, I did think of it but felt to cater for the most frequent usecase. We can definitely (based on feedback) config specific to configuring and deploying servicemonitor.

PR #224, opened 2023-09-12
feat(API)!: move deployment related config to deployment spec

sthaha:
This PR moves all deployment of kepler related specs under `deployment` field. 
I additionally cleans up the unsupported tectonic:ui tags - see https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md for supported tags


bundle/manifests/kepler-operator.clusterserviceversion.yaml
```diff
@@ -330,5 +327,5 @@ spec:
   provider:
     name: Kepler Operator Contributors
     url: https://sustainable-computing.io/
-  replaces: kepler-operator.v0.7.2
+  replaces: kepler-operator.v0.0.0-e2e
```
vimalk78: needs change

sthaha, line 330: This shouldn't be the case

sthaha: Thanks for pointing that out

vimalk78 (approved):
lgtm

PR #225, opened 2023-09-12
feat: support vanilla k8s cluster

sthaha:
This PR add a `--openshift` flag to operator which instructs operator
    to deploy components that are specific to openshift only if the flag is
    set. This allows the operator to be run against a vanilla k8s cluster by
    not setting this flag (which is the default).


sthaha:
@husky-parul this may help with running the operator against vanilla k8s. We will need a helm chart to install the operator.

husky-parul (approved):
LGTM

PR #226, opened 2023-09-12
fix(bundle): installation of bundle on openshift

sthaha:
This commit fixes the installation of bundle on openshift which fails because of the required CRD servicemonitor (which is provided by Prometheus Operator). Although the CRD is present in the cluster, OLM seems to be installing PO which conflicts with the in-cluster Prometheus Operator. The fix is to remove the dependency in CSV on ServiceMonitors


vprashar2929 (approved):
LGTM

PR #229, opened 2023-09-13
fix: restore SCC to allow privileged containers on OpenShift

sthaha:
A good reference for the SCC required is the node-exporter SCC from CMO that deploys run node-exporter daemonset. 
see:   https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/security-context-constraints.yaml 

Without this patch, although the pods are deployed, the reconciler keeps running in infinite loop.

### DIFF
--- a/pkg/controllers/kepler.go
+++ b/pkg/controllers/kepler.go
@@ -408,7 +408,13 @@ func exporterReconcilers(k *v1alpha1.Kepler, cluster k8s.Cluster) []reconciler.R
 		return rs
 	}
 
-	rs := updatersForResources(k,
+	rs := []reconciler.Reconciler{}
+	if cluster == k8s.OpenShift {
+		updater := newUpdaterForKepler(k)
+		rs = append(rs, updater(exporter.NewSCC(components.Full, k)))
+	}
+
+	rs = append(rs, updatersForResources(k,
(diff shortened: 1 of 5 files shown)


sthaha:
@BGrasnick , This PR reverts the changes made to SCC. Could you please help me understand how the changes to the SCC to make it more restrictive were tested?

sthaha:
cc: @vprashar2929

pkg/controllers/kepler.go
```diff
@@ -408,7 +408,13 @@ func exporterReconcilers(k *v1alpha1.Kepler, cluster k8s.Cluster) []reconciler.R
 		return rs
 	}
 
-	rs := updatersForResources(k,
+	rs := []reconciler.Reconciler{}
+	if cluster == k8s.OpenShift {
+		updater := newUpdaterForKepler(k)
+		rs = append(rs, updater(exporter.NewSCC(components.Full, k)))
+	}
```
sthaha, line 415: NOTE: Have the SCC created before pods are deployed.

sthaha:
@vprashar2929 and I testing this again on Openshift and the current SCC works fine. Hence closing this PR. 
Thanks a lot @vprashar2929 🙇

PR #231, opened 2023-09-13
fix(bundle): setting default channel to alpha

sthaha:
When a bundle has more than on channel annotation, it must also have the annotation for indicating the default channel, otherwise the bundle gets rejected by the community-catalog CI.

This was required for making the 0.8.0 release

See: https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/3253


vimalk78 (approved):
lgtm

PR #233, opened 2023-09-13
chore: update release workflow to include version replaces in bundle

vprashar2929:
This PR updates the release workflow to include `VERSION_REPLACED` while building the bundle. This will be useful when publishing the bundle to the community catalog.

Test workflow [run](https://github.com/vprashar2929/kepler-operator/actions/runs/6175536737/job/16762617447#step:7:54)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,12 +4,12 @@ on:
     # Inputs the workflow accepts.
     inputs:
       tag:
-        description: "Tag name, e.g. 0.4.0"
-        default: 0.6.0
+        description: "Tag name, e.g. 0.8.0"
+        default: ""
         required: true
-      release:
-        description: "Release name, e.g. release-0.4.0"
-        default: release-0.6.0
+      version-replace:
+        description: "Version to be replaced in bundle, e.g. 0.7.3"
+        default: ""
         required: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -4,12 +4,16 @@ on:
...
         required: true
       release:
-        description: "Release name, e.g. release-0.4.0"
-        default: release-0.6.0
+        description: "Release name, e.g. release-0.8.0"
+        default: release-0.8.0
+        required: true
+      replaces:
+        description: "Version to be replaced in bundle, e.g. 0.7.3"
+        default: 0.7.3
```
sthaha, line 16: Shouldn't the default be "" ?

sthaha (changes requested):
Thanks for the patch @vprashar2929. A minor change required.

.github/workflows/release.yaml
vprashar2929: Agree we should change to `""`. I think the same goes for tag and release as well right?

sthaha: thinking of it, we should only need the `tag` and the `version-replace` the `release` can be computed :)

sthaha:
@vprashar2929 , the latest patch looks good to me, could you please point us to the last test you ran in your fork with these changes applied?

vprashar2929:
Here is the latest workflow run: https://github.com/vprashar2929/kepler-operator/actions/runs/6243158617/job/16948142128

sthaha (approved):

PR #234, opened 2023-09-13
feat: add oc to make tools

vprashar2929:
This PR adds support for deploying oc with version 4.13.0 as part of `make tools`

### DIFF
--- a/hack/tools.sh
+++ b/hack/tools.sh
@@ -33,11 +33,13 @@ declare -r CONTROLLER_TOOLS_VERSION=${CONTROLLER_TOOLS_VERSION:-v0.12.1}
 declare -r OPERATOR_SDK_VERSION=${OPERATOR_SDK_VERSION:-v1.27.0}
 declare -r YQ_VERSION=${YQ_VERSION:-v4.34.2}
 declare -r CRDOC_VERSION=${CRDOC_VERSION:-v0.6.2}
+declare -r OC_VERSION=${OC_VERSION:-4.13.0}
 
 # install
 declare -r KUSTOMIZE_INSTALL_SCRIPT="https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 declare -r OPERATOR_SDK_INSTALL="https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/operator-sdk_${GOOS}_${GOARCH}"
 declare -r YQ_INSTALL="https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_${GOOS}_${GOARCH}"
+declare -r OC_URL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/$OC_VERSION"
 
 source "$PROJECT_ROOT/hack/utils.bash"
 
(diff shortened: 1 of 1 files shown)


hack/tools.sh
```diff
@@ -138,7 +140,26 @@ install_crdoc() {
 	}
 	ok "crdoc was installed successfully"
 }
+install_oc() {
```
sthaha, line 143:
```suggestion

install_oc() {
```

hack/tools.sh
```diff
@@ -138,7 +140,26 @@ install_crdoc() {
...
+	local version=" $OC_VERSION"
+	local os="$GOOS"
+	[[ $os == "darwin" ]] &&
+		os="mac"
```
sthaha, line 147:
```suggestion
	[[ $os == "darwin" ]] && os="mac"
```

hack/tools.sh
```diff
@@ -33,11 +33,13 @@ declare -r CONTROLLER_TOOLS_VERSION=${CONTROLLER_TOOLS_VERSION:-v0.12.1}
 declare -r OPERATOR_SDK_VERSION=${OPERATOR_SDK_VERSION:-v1.27.0}
 declare -r YQ_VERSION=${YQ_VERSION:-v4.34.2}
 declare -r CRDOC_VERSION=${CRDOC_VERSION:-v0.6.2}
+declare -r OC_VERSION=${OC_VERSION:-4.13.0}
 
 # install
 declare -r KUSTOMIZE_INSTALL_SCRIPT="https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 declare -r OPERATOR_SDK_INSTALL="https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/operator-sdk_${GOOS}_${GOARCH}"
 declare -r YQ_INSTALL="https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_${GOOS}_${GOARCH}"
+declare -r OC_INSTALL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/$OC_VERSION"
```
sthaha, line 42:
```suggestion
declare -r OC_URL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/$OC_VERSION"
```
 is OC_VERSION defined?

hack/tools.sh
```diff
@@ -138,7 +140,26 @@ install_crdoc() {
 	}
 	ok "crdoc was installed successfully"
 }
+install_oc() {
+	local version=" $OC_VERSION"
+	local os="$GOOS"
```
sthaha, line 145: this assume "$GOOS" is defined.  rather 

```suggestion
	local os
	os="$(go env GOOS)"
```

hack/tools.sh
```diff
@@ -138,7 +140,26 @@ install_crdoc() {
...
+	[[ $(command -v oc) ]] &&
+		[[ $(oc version --client -oyaml | grep releaseClientVersion | cut -f2 -d:) = "$version" ]] && {
+		ok "oc is already installed"
+		return 0
+	}
+	info "installing oc with version: $OC_VERSION"
+	curl -sNL "$install" | tar -xzf - -C "$LOCAL_BIN" || {
+		fail "failed to install oc"
+		return 1
+	}
```
sthaha, line 158:
```suggestion
	
	 command -v oc &&
		[[ $(oc version --client -oyaml | grep releaseClientVersion | cut -f2 -d:) == "$version" ]] && {
		ok "oc is already installed"
		return 0
	}
	
	info "installing oc with version: $OC_VERSION"
	local install=$"OC_URL/openshift-client-${os}.tar.gz"
	curl -sNL "$install" | tar -xzf - -C "$LOCAL_BIN" || {
		fail "failed to install oc"
		return 1
	}
```

sthaha (changes requested):
Could you also please add a `make target - oc`

hack/tools.sh
vprashar2929: Yes we are defining [here](https://github.com/sustainable-computing-io/kepler-operator/blob/d04ae6534e458c60a70ac97a76e8b979c5f50722/hack/tools.sh#L36)

hack/tools.sh
vprashar2929: Yes, but we already define GOOS and GOARCH above.

vprashar2929:
Why do we need to have make target for oc? It is part of make tools

sthaha:
> Why do we need to have make target for oc? It is part of make tools

See other tools target like `kustomize` `yq` etc

sthaha (approved):

PR #236, opened 2023-09-15
improve dev UX by removing unnecessary toil

sthaha:
* removes unnecessary rebases due to bundle generation
* Adds make target for installing oc
* Removes (ab)use of [[ $(command -v ...) ]] with command -v

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,10 @@ operator-sdk: ## Download operator-sdk locally if necessary.
 yq: ## Download yq locally if necessary
 	@./hack/tools.sh yq
 
+.PHONY: oc
+oc: ## Download oc locally if necessary
+	@./hack/tools.sh oc
+
 .PHONY: install-govulncheck
 install-govulncheck: ## Download govulncheck locally if necessary
 	@./hack/tools.sh govulncheck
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -265,6 +265,10 @@ operator-sdk: ## Download operator-sdk locally if necessary.
 yq: ## Download yq locally if necessary
 	@./hack/tools.sh yq
 
+.PHONY: oc
+oc: ## Download yq locally if necessary
```
vimalk78, line 269: comment says `yq`

hack/bundle.sh
```diff
@@ -101,6 +101,14 @@ main() {
 		--select-optional name=operatorhub \
 		--optional-values=k8s-version=1.25 \
 		--select-optional suite=operatorframework
+
+	local csv_file=bundle/manifests/kepler-operator.clusterserviceversion.yaml
+
+	if git diff --ignore-matching-lines='createdAt:' --exit-code "$csv_file" >/dev/null; then
+		info "no changes to $(basename $csv_file) detected; resetting it"
+		run git checkout -- "$csv_file"
```
vimalk78, line 109: is it safe to do `git checkout` from script

sthaha: yeah, the git diff explicitly checks if the only change is in createdAt ..

Makefile
sthaha: Good catch :)

vimalk78 (approved):

hack/bundle.sh
sthaha: hehe - https://github.com/operator-framework/operator-sdk/issues/6285#issuecomment-1415350333

PR #237, opened 2023-09-18
feat(test): added e2e tests for node selector and Tolerations

vimalk78:
This PR adds e2e tests for testing node selector and Tolerations.

### DIFF
--- a/pkg/utils/test/framework.go
+++ b/pkg/utils/test/framework.go
@@ -17,8 +17,10 @@ limitations under the License.
 package test
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -26,6 +28,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/sustainable.computing.io/kepler-operator/pkg/api/v1alpha1"
 	"github.com/sustainable.computing.io/kepler-operator/pkg/utils/k8s"
+	"github.com/sustainable.computing.io/kepler-operator/pkg/utils/test/oc"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -189,4 +192,44 @@ func (f Framework) WaitUntilKeplerCondition(name string, t v1alpha1.ConditionTyp
 			return condition.Status == v1alpha1.ConditionTrue, nil
 		})
 	return &k
+}
+
+func (f Framework) GetResourceNames(kind string) ([]string, error) {
+	res, err := oc.Get().Resource(kind, "").OutputJsonpath("{.items[*].metadata.name}").Run()
+	if err != nil {
+		return []string{}, err
+	}
+	nodes := strings.Split(res, " ")
+	return nodes, nil
+}
+
+func (f Framework) AddResourceLabels(kind, name string, l map[string]string) error {
+	b := new(bytes.Buffer)
+	for label, value := range l {
+		fmt.Fprintf(b, "%s=%s ", label, value)
+	}
+	return f.AddResourceLabelsStr(kind, name, b.String())
+}
+
+func (f Framework) AddResourceLabelsStr(kind, name, l string) error {
+	_, err := oc.Literal().From("oc label %s %s %s", kind, name, l).Run()
+	return err
+}
+
+func (f Framework) RemoveResourceLabels(kind, name string, l []string) error {
+	b := new(bytes.Buffer)
+	for _, label := range l {
+		fmt.Fprintf(b, "%s- ", label)
+	}
+	_, err := oc.Literal().From("oc label %s %s %s", kind, name, b.String()).Run()
+	return err
+}
+
+func (f Framework) GetTaints(node string) (string, error) {
+	return oc.Get().Resource("node", node).OutputJsonpath("{.spec.taints}").Run()
+}
+
+func (f Framework) TaintNode(node, taintStr string) error {
+	_, err := oc.Literal().From("oc adm taint node %s %s", node, taintStr).Run()
+	return err
 }
--- a/pkg/utils/test/oc/adm_top.go
+++ b/pkg/utils/test/oc/adm_top.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package oc
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Execer is interface for collecting arguments for Exec command
+type AdmTopCommand interface {
+	Command
+
+	//ForContainers to retrieve `oc adm top` of containers
+	ForContainers() AdmTopCommand
+
+	//NoHeaders to not include noHeaders
+	NoHeaders() AdmTopCommand
+}
+
+type admTop struct {
+	*runner
+	namespace  string
+	podname    string
+	containers bool
+	noHeaders  bool
+}
+
+// AdmTop creates an 'oc adm top' command
+func AdmTop(namespace, name string) AdmTopCommand {
+	e := &admTop{
+		runner:    &runner{},
+		namespace: namespace,
+		podname:   name,
+	}
+	e.collectArgsFunc = e.args
+	return e
+}
+
+func (e *admTop) ForContainers() AdmTopCommand {
+	e.containers = true
+	return e
+}
+
+func (e *admTop) NoHeaders() AdmTopCommand {
+	e.noHeaders = true
+	return e
+}
+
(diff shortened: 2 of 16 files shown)


pkg/utils/test/oc/literal.go
```diff
@@ -0,0 +1,64 @@
...
+
+	cmdstr string
+}
+
+// Literal takes an oc command string and runs it
+func Literal() ILiteral {
+	l := &literal{
+		runner: &runner{},
+	}
+	l.collectArgsFunc = l.args
```
sthaha, line 30: this feels a bit odd.. what is being done here 🤔 .

pkg/utils/test/oc/literal.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type literal struct {
+	*runner
+
+	cmdstr string
+}
+
+// Literal takes an oc command string and runs it
+func Literal() ILiteral {
```
sthaha, line 26: How about 
```suggestion
func Literal(cmd string, args ...interface{)) ILiteral {
  l.cmdstring = "oc" + strings.Trimspace(....) 
  
```
So the call would be 
```
oc.Literal("apply foobar")

File: pkg/utils/test/oc/literal_test.go [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+			scenario: "add labels to a node",
+		},
+		{
+			literal:  oc.Literal().From("oc   "),
+			occmd:    "command too small",
+			scenario: "fetch pods from namespace with labels",
+		},
+		{
+			literal:  oc.Literal().From("abc xyz"),
+			occmd:    "error: command string must start with 'oc'",
```
sthaha, line 34: how about we just add `"oc "`  for all `Literal` ?

pkg/utils/test/oc/literal.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// For oc commands not a part of this package, e.g. oc.Logs, oc.Apply, oc.Delete etc,
+// oc.Literal is a workaround to run those commands.
+
+// ILiteral is an interface for collecting the command string
+type ILiteral interface {
```
sthaha, line 12: Could we just call it Literal. 
And call `func Literal` `func Command` and it would read 
```
oc.Command("apply -f foobar.yaml).Run() 
```

pkg/utils/test/framework.go
```diff
@@ -190,3 +193,43 @@ func (f Framework) WaitUntilKeplerCondition(name string, t v1alpha1.ConditionTyp
...
+func (f Framework) AddResourceLabelsStr(kind, name, l string) error {
+	_, err := oc.Literal().From("oc label %s %s %s", kind, name, l).Run()
+	return err
+}
+
+func (f Framework) RemoveResourceLabels(kind, name string, l []string) error {
+	b := new(bytes.Buffer)
+	for _, label := range l {
+		fmt.Fprintf(b, "%s- ", label)
+	}
```
sthaha, line 223: how about using `strings.Builder` ?

vimalk78: any issues with this?

pkg/utils/test/oc/adm_top.go
```diff
@@ -0,0 +1,63 @@
+package oc
```
sthaha, line 1: missing - license :(

pkg/utils/test/oc/adm_top.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Execer is interface for collecting arguments for Exec command
+type AdmTopComman interface {
```
sthaha, line 9:
```suggestion
type AdmTopCommand interface {
```

pkg/utils/test/oc/runner.go
```diff
@@ -0,0 +1,174 @@
...
+type runner struct {
+	*osexec.Cmd
+
+	args       []string
+	configPath string
+
+	// This must be set by oc.Commands to collect arguments before calling Run
+	collectArgsFunc func() []string
+
+	tostdout bool
```
sthaha, line 41:
```suggestion
	toStdout bool
```
I feel it is better to set `output io.Writer` so that you can pass in any io.Writer including `os.Stdout and `Stderr`

vimalk78: this package is copied as-it-is from logging (path mentioned in commit message), can we treat it as a 3rd part code used only for test, and not do a line by line review?

pkg/utils/test/oc/adm_top.go
vimalk78: done

pkg/utils/test/oc/adm_top.go
vimalk78: done

pkg/utils/test/framework.go
sthaha: I don't think there is an issue with usage in test per se ... but I would heed - https://github.com/golang/go/blob/master/src/bytes/buffer.go#L65

pkg/utils/test/oc/runner.go
vimalk78: changed

pkg/utils/test/oc/get.go
```diff
@@ -0,0 +1,117 @@
+package oc
```
sthaha, line 1: @vimalk78 , we need license on all our files :(

vimalk78: adding

pkg/utils/test/oc/runner.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"bytes"
+	"fmt"
+	"os"
+	osexec "os/exec"
+	"strings"
+	"time"
+
+	//logger "github.com/ViaQ/logerr/v2/log"
+	//log "github.com/ViaQ/logerr/v2/log/static"
```
sthaha, line 12: cleanup ?

vimalk78: done

pkg/utils/test/oc/get.go
vimalk78: old files also do not have license

sthaha: We need a checker in CI to do these mundane checks 🤯 ...

tests/e2e/kepler_test.go
```diff
@@ -71,3 +73,122 @@ func TestBadKepler_Reconciliation(t *testing.T) {
...
+
+	f.AssertResourceExists(components.Namespace, "", &corev1.Namespace{})
+	ds := appsv1.DaemonSet{}
+	f.AssertResourceExists(exporter.DaemonSetName, components.Namespace, &ds)
+
+	kepler := f.WaitUntilKeplerCondition("kepler", v1alpha1.Reconciled)
+	reconciled, err := k8s.FindCondition(kepler.Status.Conditions, v1alpha1.Reconciled)
+	assert.NoError(t, err, "unable to get reconciled condition")
+	assert.Equal(t, reconciled.ObservedGeneration, kepler.Generation)
+	assert.Equal(t, reconciled.Status, v1alpha1.ConditionTrue)
```
sthaha, line 105: We only need to check for availability - right? 
```suggestion
```

tests/e2e/kepler_test.go
```diff
@@ -71,3 +73,122 @@ func TestBadKepler_Reconciliation(t *testing.T) {
...
+	node := nodes[0]
+	var labels k8s.StringMap = map[string]string{"e2e-test": "true"}
+	err = f.AddResourceLabels("node", node, labels)
+	assert.NoError(t, err, "could not label node")
+
+	f.CreateKepler("kepler", func(k *v1alpha1.Kepler) {
+		k.Spec.Exporter.Deployment.NodeSelector = labels.Merge(k8s.StringMap{
+			"kubernetes.io/os": "linux",
+		})
+	})
```
sthaha, line 95: We shouldn't need to add the `kubernetes.io/os: linux` label since the operator is expected to set it.

tests/e2e/kepler_test.go
```diff
@@ -71,3 +73,122 @@ func TestBadKepler_Reconciliation(t *testing.T) {
...
+	kepler = f.WaitUntilKeplerCondition("kepler", v1alpha1.Available)
+	assert.EqualValues(t, 1, kepler.Status.NumberAvailable)
+
+	f.DeleteKepler("kepler")
+
+	ns := components.NewKeplerNamespace()
+	f.AssertNoResourceExists(ns.Name, "", ns)
+	f.AssertNoResourceExists(exporter.DaemonSetName, components.Namespace, &ds)
+
+	err = f.RemoveResourceLabels("node", node, []string{"e2e-test"})
```
sthaha, line 116: we should add cleanup (`t.Cleanup()`) in the same method that Initialise the resource labels.. See `CreateKepler` that automatically deletes it after the test is over.  Otherwise, a test failure can result in unrelated tests failing.

vimalk78: changed

tests/e2e/kepler_test.go
vimalk78: changed

tests/e2e/kepler_test.go
vimalk78: changed

sthaha (approved):

tests/e2e/kepler_test.go
sthaha: I think `f.AddResourceLabels()` should add the `f.T.Cleanup()` like `CreateKepler` calls `DeleteKepler`. The idea with these methods are that the caller should not be worried about cleanup. 

I will submit a followup patch.

PR #238, opened 2023-09-18
feat: package go test to run in prow CI

vprashar2929:
This PR creates a Dockerfile under the `tests` dir which will build a test binary that downstream prow CI will consume in running e2e tests on OpenShift.

### DIFF
--- a/.github/publish-images/action.yaml
+++ b/.github/publish-images/action.yaml
@@ -1,5 +1,5 @@
 name: Build and Publish Images
-description: 'Publishes operator and bundle images to an Image Registry'
+description: "Publishes operator and bundle images to an Image Registry"
 inputs:
   image_registry:
     description: "image registry"
@@ -30,14 +30,13 @@ runs:
         check-latest: true
         cache: true
 
-    - name: Login to Image Registry 
+    - name: Login to Image Registry
       uses: docker/login-action@v2
       if: "!startsWith(inputs.image_registry, 'localhost')"
       with:
         registry: ${{ inputs.image_registry }}
         username: ${{ inputs.registry_login }}
         password: ${{ inputs.registry_token }}
-      
 
     - name: Build Operator
       shell: bash
@@ -56,6 +55,14 @@ runs:
         VERSION: ${{ inputs.version }}
         IMG_BASE: ${{ inputs.image_registry }}
         ADDITIONAL_TAGS: ${{ inputs.additional_tags }}
+
+    - name: Build E2E
+      shell: bash
+      run: |
+        make e2e-test-image
+      env:
+        VERSION: ${{ inputs.version }}
+        IMG_BASE: ${{ inputs.image_registry }}
 
     - name: Push Images
       shell: bash
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ ADDITIONAL_TAGS ?=
 
 KEPLER_IMG ?= $(IMG_BASE)/kepler:$(KEPLER_VERSION)
 
+# E2E_TEST_IMG defines the image:tag used for the e2e test image
+E2E_TEST_IMG ?=$(IMG_BASE)/kepler-operator-e2e:$(VERSION)
 
 .PHONY: fresh
 fresh: ## default target - sets up a k8s cluster with images ready for deployment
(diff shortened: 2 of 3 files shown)


tests/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# The Dockerfile's resulting image is used in executing Kepler Operator e2e tests on ProwCI
```
sthaha, line 1: So there is no escaping this :) ?

sthaha: @vprashar2929 could you please also 
* add a make target - `e2e-image`  that follows the same logic as`operator-build` 
* test the image can be built by adding this to the build-images jobs ?

sthaha (changes requested):
Need to add CI validation that the image can be built. 

Do we also need to publish the image?

tests/Dockerfile
vprashar2929: Yes definitely need that

vprashar2929:
> Do we also need to publish the image?

No we don't need to publish the image. Prow CI pipeline will build the image and push to its registry.

.github/publish-images/action.yaml
```diff
@@ -1,5 +1,5 @@
 name: Build and Publish Images
-description: 'Publishes operator and bundle images to an Image Registry'
+description: "Publishes operator and bundle images to an Image Registry"
```
sthaha, line 2: nit: "" are optional :)

tests/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+FROM golang:1.20
+
+WORKDIR /workspace
+
+RUN apt-get update
```
sthaha, line 6: I think we can remove this since we aren't installing any package. Having this step has the downside that the image is unnecessarily bloated and the layer can't be cached.

sthaha, line 6:
```suggestion

```

sthaha (changes requested):
Looks good except for one small change in Dockerfile.

Makefile
```diff
@@ -40,6 +40,8 @@ ADDITIONAL_TAGS ?=
 
 KEPLER_IMG ?= $(IMG_BASE)/kepler:$(KEPLER_VERSION)
 
+# E2E_TEST_IMG defines the image:tag used for the e2e test image
+E2E_TEST_IMG ?=$(IMG_BASE)/kepler-operator-e2e:$(VERSION)
```
sthaha, line 44: I have created - https://quay.io/repository/sustainable_computing_io/kepler-operator-e2e 
@rootfs do we need make additional changes for the bots to be able to push to this repo?

vprashar2929: @sthaha Are we also planning to use the e2e image outside Prow CI? Regarding Prow CI usage, we don't need to push the e2e image to any repo. It will build the image and push it to its internal registry

sthaha: How is the image pushed?

vprashar2929: Just like how bundle and operator images are built and pushed. [Here](https://github.com/openshift/release/blob/2644055d79d1ae8172280197e1cb281cca9cd90e/ci-operator/config/vprashar2929/kepler-operator/vprashar2929-kepler-operator-package.yaml#L25) we define the list of images to build and in the test's [steps](https://github.com/openshift/release/blob/2644055d79d1ae8172280197e1cb281cca9cd90e/ci-operator/step-registry/vprashar2929/tests/kepler/vprashar2929-tests-kepler-ref.yaml#L3) we make use of the build image

.github/publish-images/action.yaml
```diff
@@ -57,6 +56,14 @@ runs:
...
         ADDITIONAL_TAGS: ${{ inputs.additional_tags }}
 
+    - name: Build E2E
+      shell: bash
+      run: |
+        make e2e-test-image
+      env:
+        VERSION: ${{ inputs.version }}
+        IMG_BASE: ${{ inputs.image_registry }}
+
```
sthaha, line 66: @vprashar2929 why are we buidling the image if that isn't being used anywhere in CI or need to be pushed? 🤔 . Is to to ensure the image can be built?

tests/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Install kubectl and oc
+RUN curl -L -o oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/openshift-client-linux.tar.gz \
+    && tar -xvzf oc.tar.gz \
+    && chmod +x kubectl oc \
+    && mv oc kubectl /usr/local/bin/
```
sthaha, line 12: Moving this after  `go mod download` may speedup (re)builds.

.github/publish-images/action.yaml
vprashar2929: Yeah exactly.

sthaha: could you please add a note 🙏 ?

sthaha (changes requested):
Some minor comments to address.

sthaha (approved):

PR #239, opened 2023-09-19
feat(api): add KeplerExperiment api to support experimental features

sthaha:
The idea behind this PR is to have a low barrier for entry to add features to the operator which affecting the stable API/ release. The experimental features are enabled by `make <target> EXPERIMENT=true`. 

This allows us to iterate quickly (instead of bike shedding) on API and to learn from real world scenarios. 

In this PR, a `KeplerExperiment` CRD is added which replicates the same controller logic for deploying 
Kepler. 

## Note for  changes in future  ## 
* all experimental features should only be enabled when `build tag` experiment is defined
* when adding an experimental feature / api - `make bundle` should not reflect any change.
* All experimental API should not be part of the bundle 
* All experimental controllers should be a NO-OP when running in production mode


sthaha:
cc: @vimalk78 @rootfs

sunya-ch:
Thank you for opening this route.

I have a few points for discussion.
1. What is the benefit of adding new CR with duplicated crds comparing to testing with operatorsdk run customized bundle image?
2. How to migrate from Experimental feature (new API/crd changes) to main one (single make command run or manually copy/paste)? What is the condition/procedure and what are steps to do? Should it be the same API version or new API version?
3. Why we don't just separate branch and channel for experimental feature and merge when condition in 2. is satisfied? We might consider having pre-alpha/alpha/beta/stable channels.

husky-parul:
@sthaha Can you please elaborate on the following points?

> all experimental features should only be enabled when build tag experiment is defined
> have we used `operator-sdk create` to create the scaffolding and how we differentiate in the CI between the testing of experimental feature and existing v1alpha1?
> migration scheme from experimental to v1alpha1 (@sunya-ch also raised this point)

vimalk78:
are we anticipating experiments which perhaps will not graduate to mainline? if it is only for incorporating remaining kepler components, we can do it by branching or feature-flags.

sthaha:
@sunya-ch @vimalk78 @husky-parul , sorry for not getting back to this early. 
Could we please start a discussion threads for each of the points raised and address each individually? 

I have created some already. 
See: https://github.com/sustainable-computing-io/kepler-operator/discussions/categories/general

PR #240, opened 2023-09-20
style: Add info related to uwm in Grafana script

vprashar2929:
This PR adds the info inside `deploy-grafana.sh` related to how to do custom configuring of Prometheus deployed by uwm according to requirement.


hack/dashboard/openshift/deploy-grafana.sh
```diff
@@ -329,6 +344,7 @@ main() {
 	[[ -z "$(ls -A "$BACKUP_DIR")" ]] && rm -rf "$BACKUP_DIR"
 
 	show_key_info
+	show_uwm_info
```
sthaha, line 347: I would have added this to the `show_key_info` `key` as in important :)

sthaha (approved):

PR #242, opened 2023-09-20
feat(bundle): Use user supplied kepler image

vimalk78:
For dev testing sometimes it is needed to change the kepler image to be deployed. This PR lets devs use an env variable to specify the kepler image.

 - Makefile: variable(s) for defining kepler image, which form the 'official' image used in the bundle, but whose value can be overridden by user while building bundle
 - CSV: added env variable `RELATED_IMAGE_KEPLER` in kepler operator deployment which gets passed as an argument to manager, and its value can be overridden by user while building the bundle by supplying env variables `KEPLER_IMG` or `KEPLER_VERSION`
 - exporter/main/go: changed to use the new env variable `RELATED_IMAGE_KEPLER`
 - hack/bundle.sh: replace `KEPLER_IMG` with its value in CSV
 - operator-sdk adds `relatedImages` field based on `RELATED_IMAGE*` variables used

**Example Usage 1**: use your own built kepler image from `quay.io/<your-repo>`
make bundle as usual, but with extra env variable for KEPLER_VERSION. this will use kepler image from `quay.io/<your-repo>`. Then run the bundle, and create kepler instance
```
make bundle bundle-build bundle-push IMG_BASE=quay.io/vimalkum VERSION=0.0.0-test-1  KEPLER_VERSION=latest-libbpf

operator-sdk run bundle quay.io/vimalkum/kepler-operator-bundle:0.0.0-test-1  --install-mode AllNamespaces --namespace=operators

oc apply -f ./config/samples/kepler.system_v1alpha1_kepler.yaml
```

This will run kepler with image `quay.io/vimalkum/kepler:latest-libbpf`

**Example Usage 2:** use kepler image from another repo, e.g. `quay.io/sustainability_computing_io`
make bundle as usual, but with extra env variable for KEPLER_IMG. this will use kepler image as per variable. Then run the bundle, and create kepler instance
```
make bundle bundle-build bundle-push IMG_BASE=quay.io/vimalkum VERSION=0.0.0-test-1  KEPLER_IMG=quay.io/sustainable_computing_io/kepler:latest-libbpf

operator-sdk run bundle quay.io/vimalkum/kepler-operator-bundle:0.0.0-test-1  --install-mode AllNamespaces --namespace=operators

oc apply -f ./config/samples/kepler.system_v1alpha1_kepler.yaml
```

This will run kepler with image `quay.io/sustainable_computing_io/kepler:latest-libbpf`

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -59,14 +59,6 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
-func env(name, defaultValue string) string {
-	value := os.Getenv(name)
-	if len(value) == 0 {
-		return defaultValue
-	}
-	return value
-}
-
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
@@ -82,8 +74,8 @@ func main() {
 	flag.BoolVar(&openshift, "openshift", false,
 		"Indicate if the operator is running on an OpenShift cluster.")
 
-	// NOTE: KEPLER_IMAGE can be set as env or flag, flag takes precedence over env
-	keplerImage := env("KEPLER_IMAGE", exporter.StableImage)
+	// NOTE: RELATED_IMAGE_KEPLER can be set as env or flag, flag takes precedence over env
+	keplerImage := os.Getenv("RELATED_IMAGE_KEPLER")
 	flag.StringVar(&exporter.Config.Image, "kepler.image", keplerImage, "kepler image")
 
 	opts := zap.Options{
--- a/tests/run-e2e.sh
+++ b/tests/run-e2e.sh
@@ -8,6 +8,7 @@ declare -r PROJECT_ROOT
 
 source "$PROJECT_ROOT/hack/utils.bash"
 
+declare -r LOCAL_BIN="$PROJECT_ROOT/tmp/bin"
 declare -r OPERATOR="kepler-operator"
 declare -r OLM_CATALOG="kepler-operator-catalog"
 declare -r VERSION="0.0.0-e2e"
@@ -260,7 +261,7 @@ kind_load_images() {
 	while read -r img; do
 		run docker pull "$img"
 		run kind load docker-image "$img"
-	done < <(grep -e '=\s\+"quay.io' -r pkg/components/exporter | cut -f2 -d'"')
+  done < <(yq -r .spec.relatedImages[].image bundle/manifests//kepler-operator.clusterserviceversion.yaml)
 }
 
 deploy_operator() {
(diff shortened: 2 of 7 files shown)


cmd/manager/main.go
```diff
@@ -83,7 +83,7 @@ func main() {
 		"Indicate if the operator is running on an OpenShift cluster.")
 
 	// NOTE: KEPLER_IMAGE can be set as env or flag, flag takes precedence over env
-	keplerImage := env("KEPLER_IMAGE", exporter.StableImage)
+	keplerImage := env("RELATED_IMAGE_KEPLER", exporter.StableImage)
```
sthaha, line 86: is it possible to we use `KEPLER_IMAGE` itself? If it should be `RELATED_IMAGE_KEPLER` then the comment above has to change as well

sthaha (changes requested):
A minor comment to fix 
BTW, did you consider just using `-ldflags` to set the StabeImage` ? 

go build -ldflags "-X ./pkg/components/exporter.StableImage=$(KEPLER_IMAGE)"

vimalk78:
> A minor comment to fix BTW, did you consider just using `-ldflags` to set the StabeImage` ?
> 
> go build -ldflags "-X ./pkg/components/exporter.StableImage=$(KEPLER_IMAGE)"

thats a good idea, we dont have to change the version in two places. 
But perhaps we can remove this variable now and use the env variable itself to know the kepler image to be used, which will be supplied by CSV deployment.

cmd/manager/main.go
vimalk78: > is it possible to we use `KEPLER_IMAGE` itself? If it should be `RELATED_IMAGE_KEPLER` then the comment above has to change as well

`RELATED_IMAGE_`  should be the way to refer to operand images in CSV
https://docs.openshift.com/container-platform/4.13/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs

In the downstream CSV, this is a requirement, and a `relatedImages` section is automatically added in the process. just like we have got it generated here.

vimalk78: will change the comment

sthaha:
Yes, let just use the relatedImages - 2 birds in one stone :)

sthaha:
Yes, let just use the relatedImages - 2 birds in one stone :)

tests/run-e2e.sh
```diff
@@ -346,6 +347,7 @@ print_config() {
 }
 
 main() {
+	export PATH="$LOCAL_BIN:$PATH"
```
sthaha, line 350: nice!

sthaha (approved):

PR #248, opened 2023-09-26
feat: improve uninstall script to find the installed version

sthaha:
This commit improves the experience of using the uninstall script by
automatically finding the version of kepler-operator installed. So, user
need not explicitly specify the version to installed (but they can
should they choose to) to uninstall.

It also adds a --help option to the script to print the usage

### DIFF
--- a/hack/uninstall-operator.sh
+++ b/hack/uninstall-operator.sh
@@ -108,6 +108,19 @@ find_installed_version() {
 	return 0
 }
 
+list_olm_resources() {
+	local operator="$1"
+	shift
+
+	header "Listing Resources of Kepler Operator"
+
+	info_run kubectl get csv -n "$OPERATORS_NS"
+	kubectl get csv -n "$OPERATORS_NS" | grep -E "$operator|NAME" || true
+
+	run kubectl get olm -n "$OPERATORS_NS" -o wide || true
+
+}
+
 main() {
 	local operator="kepler-operator"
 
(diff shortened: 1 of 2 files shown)


sthaha:
cc: @vimalk78 @vprashar2929

vimalk78:
can we add an option to find resources even if `find_installed_version` fails? for those clusters where operator installation fails halfway.

hack/uninstall-operator.sh
```diff
@@ -87,11 +116,21 @@ main() {
...
+		return 0
+	}
+
+	[[ -z "$OPERATOR_VERSION" ]] && {
+		info "No operator version specified; finding the installed version"
+		find_installed_version "$operator"
+		ok "Found - Kepler Operator version: $OPERATOR_VERSION"
+	}
 
 	header "Resources of Kepler Operator - $OPERATOR_VERSION"
```
vimalk78, line 130: add a heavy line below printing CSV.

vprashar2929 (dismissed):
LGTM

hack/uninstall-operator.sh
sthaha: I will make it a header

sthaha: Also notices that the `operators` is lingering which I have now deleted.

vimalk78 (approved):
lgtm

PR #256, opened 2023-09-27
test: Add error check during cleanup

vprashar2929:
This PR adds the check for any error in the case when cleanup happens.

### DIFF
--- a/tests/e2e/kepler_test.go
+++ b/tests/e2e/kepler_test.go
@@ -2,7 +2,6 @@ package e2e
 
 import (
 	"encoding/json"
-	"fmt"
 	"testing"
 	"time"
 
@@ -26,7 +25,12 @@ func TestKepler_Deletion(t *testing.T) {
 
 	//
 	ds := appsv1.DaemonSet{}
-	f.AssertResourceExists(exporter.DaemonSetName, components.Namespace, &ds, test.Timeout(10*time.Second))
+	f.AssertResourceExists(
+		exporter.DaemonSetName,
+		components.Namespace,
+		&ds,
+		test.Timeout(10*time.Second),
+	)
 
 	f.DeleteKepler("kepler")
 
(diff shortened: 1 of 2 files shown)


tests/e2e/kepler_test.go
```diff
@@ -88,7 +93,8 @@ func TestNodeSelector(t *testing.T) {
 	err = f.AddResourceLabels("node", node, labels)
 	assert.NoError(t, err, "could not label node")
 	t.Cleanup(func() {
```
sthaha, line 95: @vprashar2929 thanks for the patch. Since you are at it,  could you also please  move `t.Cleanup` to `f.AddResourceLabels` itself.  There are other functions in the framework that automatically cleans up after itself so that the caller need not worry about cleaning up.

tests/e2e/kepler_test.go
```diff
@@ -131,7 +133,8 @@ func TestTaint_WithToleration(t *testing.T) {
 	assert.NoError(t, err, "failed to taint node %s", node)
 	t.Cleanup(func() {
 		// remove taint
-		f.TaintNode(node, fmt.Sprintf("%s-", e2eTestTaint.Key))
+		err := f.TaintNode(node, fmt.Sprintf("%s-", e2eTestTaint.Key))
+		assert.NoError(t, err, "could not remove taint from node")
 	})
```
sthaha, line 138: Thanks a lot for the fix above.. can we do the same here 🙏 ?

sthaha: @vprashar2929 may be in a followup PR?

sthaha (approved):

tests/e2e/kepler_test.go
vprashar2929: Completely forgot this😅 I will fix this as well in the PR

sthaha (approved):

PR #262, opened 2023-10-02
ci: add shellcheck linter

sthaha:
* adds a new CI job to run shellcheck linter
* fixes shellcheck errors and removes unused scripts


sthaha:
cc: @vprashar2929

vprashar2929:
@sthaha Can you add the description for the PR?🙂

vprashar2929:
Shouldn't we add a shellcheck for source [here](https://github.com/sustainable-computing-io/kepler-operator/blob/391d0aee01bad10ffff3d2b2e990990073114a76/tests/run-e2e.sh#L9) as well?

sthaha:
> Shouldn't we add a shellcheck for source [here](https://github.com/sustainable-computing-io/kepler-operator/blob/391d0aee01bad10ffff3d2b2e990990073114a76/tests/run-e2e.sh#L9) as well?

Somehow shellchecker is happy. I could probably be the SHELLCHECK_OPT -x that is automatically following the sourced file.

vprashar2929:
Wired🤔 then why not the same for other scripts?

PR #263, opened 2023-10-03
fix: promql used in openshift dashboards

sthaha:
This commits brings in the fixes added to kepler repo by PR: https://github.com/sustainable-computing-io/kepler/pull/948

Following changes are also made
* remove repeated namespace from all legends
* make Overview as the first item in the dropdown list by adding a space before ` O`verview


vprashar2929 (approved):
LGTM

PR #266, opened 2023-10-04
feat(dashboard): use prometheus recording rules faster queries

sthaha:
This commit improves the rendering of dashboards by taking advantage of Prometheus recording rules to pre-compute queries which may otherwise take too much memory (resulting in Prometheus OOM crash) or time for the query to execute.

The Operator deploy a `PrometheusRule` resource which pre-computes queries used in openshift dashboards.

image

image

### DIFF
--- a/pkg/components/exporter/exporter.go
+++ b/pkg/components/exporter/exporter.go
@@ -56,6 +56,8 @@ const (
 	overviewDashboardName = "power-monitoring-overview"
 	nsInfoDashboardName   = "power-monitoring-by-ns"
 	DashboardNs           = "openshift-config-managed"
+
+	PrometheusRuleName = prefix + "prom-rules"
 )
 
 // Config that will be set from outside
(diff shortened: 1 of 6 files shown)


sthaha:
@vimalk78 @vprashar2929 to make the validation easier, I have pushed a bundle created from this PR to 
` quay.io/sthaha/kepler-operator-bundle:0.0.0-dev`

sthaha:
```sh
./tmp/bin/operator-sdk run bundle quay.io/sthaha/kepler-operator-bundle:0.0.0-dev --install-mode AllNamespaces --namespace openshift-operators --skip-tls

```

pkg/components/exporter/exporter.go
```diff
@@ -509,3 +511,93 @@ func NewServiceMonitor() *monv1.ServiceMonitor {
...
+					),
+
+					record("kepler:container_other_watts:1m:by_ns_pod",
+						`sum by (container_namespace, pod_name) (
+								irate(kepler_container_other_joules_total[1m])
+						)`,
+					),
+
+					record("kepler:container_dram_watts:1m:by_ns_pod",
+						`sum by (container_namespace, pod_name) (
```
vprashar2929, line 588: Just a small question is there any reason to why we are doing sum by `container_namespace`? I see in the queries we are only doing sum by `pod_name`.

sthaha: The query also needs `container_namespace`

vprashar2929 (approved):
LGTM

PR #272, opened 2023-10-11
fix(kind): ensure prometheus can discover all services

sthaha:
Previously, prometheus deployed to monitoring namespace wasn't able to discover kepler deployed to `openshift-kepler-operator` because it lacked the rbac to discover services outside `monitoring` namespace. This commit fixes it by adding additional rbac rules allowing all services to be discovered.


vprashar2929 (approved):
LGTM

Issue #186, opened 2023-08-30
kubectl get kepler should print  default tolerations as well


sthaha:
https://github.com/sustainable-computing-io/kepler-operator/pull/182#discussion_r1309761232

PR #273, opened 2023-10-11
fix(api): define default toleration in Kepler CRD

sthaha:
Previously, `kubectl get kepler -o wide` did not show the default toleration applied viz. {"operator": "Exists"}

This PR fixes it by moving the default toleration to the CRD spec itself

Fixes: https://github.com/sustainable-computing-io/kepler-operator/issues/186


vimalk78 (approved):
lgtm

vimalk78 (approved):
lgtm

PR #276, opened 2023-10-13
fix(exporter): do no enable estimated idle power by default

sthaha:
The 'idle' value refers to the total host idle power, which is not distributed among all the running VMs on the node. Consequently, each VM may wrongly double-count the host's idle time if there are multiple VMs on the same host.

In public cloud environments, it is impractical to determine the number of running VMs on a node. This parameter is intended for use only one bare metal servers that rely on node power estimation because they do not provide real-time power metrics, or when there is only one VM running on the host for very specific scenarios.

So, lets enable this setting through Kepler spec (new feature).

Ref: https://github.com/sustainable-computing-io/kepler/issues/973#issuecomment-1760680200


vprashar2929 (approved):
LGTM

Issue #203, opened 2023-09-04
Cache commonly used binaries in CI


sthaha:
GH CI is starting to fail dowloading binaries 
See: https://github.com/sustainable-computing-io/kepler-operator/actions/runs/6071218125/job/16468811589?pr=201#step:4:142
```
Flag --short has been deprecated, and will be removed in the future.
 🔔 installing kustomize version: v3.8.7
Github rate-limiter failed the request. Either authenticate or wait a couple of minutes.
 ❌ FAIL: failed to install kustomize
make: *** [Makefile:201: kustomize] Error 1
```

We don't need to download binaries every time if we cache them. E.g https://github.com/rhobs/observability-operator/blob/main/.github/tools-cache/action.yaml 

This should also improve both the stability and  the time taken by CI as well.

sthaha:
@sbdtu5498 thank you for helping you 🙇 . Please note we already have `make tools` to help with the installation of required binaries to `tmp/bin`

sthaha:
Reassigning this to @vprashar2929 since we have #278  from him.

vprashar2929:
Closing as #282 is merged

PR #278, opened 2023-10-16
feat(ci): add support to cache tools in CI

vprashar2929:
This PR fixes issue #203 

CI runs:

Cache: https://github.com/vprashar2929/kepler-operator/actions/caches

Install tools: https://github.com/vprashar2929/kepler-operator/actions/runs/6529594381/job/17727500939#step:4:6

Sample run: https://github.com/vprashar2929/kepler-operator/actions/runs/6529594381

### DIFF
--- a/.github/publish-images/action.yaml
+++ b/.github/publish-images/action.yaml
@@ -38,6 +38,9 @@ runs:
         username: ${{ inputs.registry_login }}
         password: ${{ inputs.registry_token }}
 
+    - name: Install all tools
+      uses: ./.github/tools-cache
+
     - name: Build Operator
       shell: bash
       run: |
--- a/.github/tools-cache/action.yaml
+++ b/.github/tools-cache/action.yaml
@@ -1,0 +1,21 @@
+---
+name: tools-cache
+description: Caches development tools
+runs:
+  using: composite
+  steps:
+    - uses: actions/cache@v3
+      id: tools-cache
+      with:
+        path: ./tmp/bin
+        key: ${{ runner.os }}-tools-${{ hashFiles('./.github/tools-cache') }}
+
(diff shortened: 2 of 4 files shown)


.github/publish-images/action.yaml
```diff
@@ -38,6 +38,9 @@ runs:
         username: ${{ inputs.registry_login }}
         password: ${{ inputs.registry_token }}
 
+    - name: Install all tools
+      uses: ./.github/tools-cache
```
sthaha, line 42: Just a thought, perhaps we should move all actions to `actions`

.github/tools-cache/action.yaml
```diff
@@ -0,0 +1,21 @@
...
+name: tools-cache
+description: Caches development tools
+runs:
+  using: composite
+  steps:
+    - uses: actions/cache@v3
+      id: tools-cache
+      with:
+        path: ./tmp/bin
+        key: ${{ runner.os }}-tools-${{ hashFiles('./.github/tools-cache') }}
```
sthaha, line 11: We need to may be hash `hack/tools.sh` instead. What ObO does is to create a '.github/tools` file see: https://github.com/rhobs/observability-operator/blob/main/.github/tools
The hash of that file is used to check of new binaries need to be downloaded.

sthaha (changes requested):
We need a file that can be hashed. I think using `hacks/tools.sh` may be sufficient even if that means any changes to `tools.sh` will result in binaries bring downloaded again. Or make tools.sh write the `.github/tools` files like ObO

sthaha:
@vprashar2929 this got closed automatically because the branch was deleted. Can you please create a PR from your fork?

PR #281, opened 2023-10-18
fix(deployment): do not use generic k8s labels

sthaha:
Previously, deploying operator created deployment that used generic
`control-plane: controller-manager` label. This causes external tools
(including the must-gather script) to lookup pods that can belong to
**any** deployment that has this label. This commit replaces the label
with `app.kubernetes.io/instance: controller-manager`.

Fixes: https://issues.redhat.com/browse/POWERMON-159


vimalk78 (approved):
lgtm

PR #285, opened 2023-10-19
chore: ensure all scripts are formatted the same

sthaha:
This PR makes the following changes
* Adds `shfmt` tool to format shell scripts
* Formats all shellscripts using `shfmt`
* Modifies CI to use `make fmt` target instead of the shell script so that changes can be checked locally


hack/tools.sh
```diff
@@ -123,6 +124,42 @@ install_yq() {
...
+	GOBIN=$LOCAL_BIN \
+		go install "$pkg@$version" || {
+		fail "failed to install $pkg - $version"
+		return 1
+	}
+	ok "$pkg - $version was installed successfully"
+
+}
+
+validate_verison() {
```
vprashar2929, line 143: Love the `validate_version` addition. Just a thought since we have it here maybe we can update other tools to make use of it as well?

sthaha: yes, I thought of it but I felt it is a feature creep. And it can be done is a different PR.

vprashar2929 (approved):
LGTM

Issue #195, opened 2023-08-31
Make it possible to adjust more fields in spec like image repository or resources


BGrasnick:
Hey,

as of right now, we can only specify the port, nodeSelectors and Tolerations for the kepler Pod when using the kepler-operator.
I think it would be great to also be able to adjust the image repository for internal registries and the resources of the containers to help with scheduling.

The question is whether it is better to step by step add possibilities to adjust all the fields that users might want to adjust or go for a more general approach by adding support for PodSpec [like they did for the GitLab operator for example](https://gitlab.com/gitlab-org/gl-openshift/gitlab-runner-operator/-/merge_requests/137#b94564a32837cb4b9f54c17f46be72ac2cef3eef).

What do you think?

I would be happy to help and contribute but first it would be good to decide on an approach.

sthaha:
> I think it would be great to also be able to adjust the image repository for internal registries and the resources of the containers to help with scheduling.

The reason we pin the version of kepler in operator is to 
* make it hard for the user do not run arbitrary image off the internet with with highest privilege granted to it :) 
* Allow operator to embed information about how to  configure the image that is deployed and in future, migrate gracefully from the previous version of kepler deployed.

For testing and development purpose, a new bundle can be generated pointing to any image using `--kepler.image` flag or `KEPLER_IMAGE` env var.

With the podSpec, what parts are you specifically targeting? Do you have an idea about how the spec would look? 
```
  spec:
    exporter:
        < changes > 
```

Please note that although, the operator is only at `Level - 1: Basic Install`, it would be nice if we didn't have to 
introduce changes that we may have to revert later on (breaking changes). In my experience it is better to rely on inputs from the users before adding a feature.

BGrasnick:
Hey @sthaha, thank you for your reply!
Of course I understand that you want to pin the version to make sure everything works as intended! I was rather talking about the repository where the image is coming from. We and probably other users of kepler-operator (at least in the future) would love if we could get the image from an internal registry that mirrors quay. 
PodSpec would make the whole operator very flexible. Things might break if users change the wrong parameters like the image tag but I think that is a risk they know about if they change that.
What I would love as of today would be the following:

```
  spec:
    exporter:
        image:
          repository: <internal-registry>
        resources:
          requests:
            memory: "<XYZ>Mi"
            cpu: "<XYZ>m"
          limits:
            memory: "<XYZ>Mi"
            cpu: "<XYZ>m"
```

This would then lead to using the image:

[...]

PR #287, opened 2023-10-23
Feat: add method field to exporter spec

BGrasnick:
Addresses #195.

Adds a new enum to kepler exporter spec that allows choosing which method/image to use: bcc or libbpf.

Needs to be changed once kepler 0.7.0 is released and default will be libbpf.

Testing was a little challenging and I am not super happy with the solution and open for suggestions.
As of now, the RELATED_IMAGE_KEPLER var gets set for Config.Image and determines the image used.
This is done in main.go lines 78 & 79. As this was needed for the test, I added lines 111-113 in exporter.go to set a default in case Config.Image is empty.
However, because of this and the test, now the image & its version is hardcoded in 3 different locations which is not good.
I am open to any help here :)

Was tested manually on an OpenShift 4.11 cluster.

@sthaha: any thoughts?

### DIFF
--- a/config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
+++ b/config/crd/bases/kepler.system.sustainable.computing.io_keplers.yaml
@@ -129,6 +129,12 @@ spec:
                           type: object
                         type: array
                     type: object
+                  method:
+                    default: bcc
+                    enum:
+                    - bcc
+                    - libbpf
+                    type: string
                 type: object
             type: object
           status:
--- a/config/samples/kepler.system_v1alpha1_kepler.yaml
+++ b/config/samples/kepler.system_v1alpha1_kepler.yaml
@@ -10,4 +10,5 @@ spec:
   exporter:
     deployment:
       port: 9103
+    method: libbpf # bcc | libbpf
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -117,6 +117,16 @@ KeplerSpec defines the desired state of Kepler
           <br/>
         </td>
         <td>false</td>
+      </tr><tr>
+        <td><b>method</b></td>
+        <td>enum</td>
+        <td>
+          <br/>
+          <br/>
+            <i>Enum</i>: bcc, libbpf<br/>
+            <i>Default</i>: bcc<br/>
+        </td>
+        <td>false</td>
       </tr></tbody>
 </table>
 
--- a/pkg/api/v1alpha1/kepler_types.go
+++ b/pkg/api/v1alpha1/kepler_types.go
@@ -111,6 +111,11 @@ type ExporterDeploymentSpec struct {
 
 type ExporterSpec struct {
 	Deployment ExporterDeploymentSpec `json:"deployment,omitempty"`
+
+	// +kubebuilder:default="bcc"
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Enum={"bcc","libbpf"}
+	Method string `json:"method,omitempty"`
 }
 
 // KeplerSpec defines the desired state of Kepler
--- a/pkg/components/exporter/exporter.go
+++ b/pkg/components/exporter/exporter.go
@@ -106,6 +106,17 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.Kepler) *appsv1.DaemonSe
 
 	bindAddress := "0.0.0.0:" + strconv.Itoa(int(deployment.Port))
 
+	configImage := Config.Image
+
+	if configImage == "" {
+		configImage = "quay.io/sustainable_computing_io/kepler:release-0.6.1"
+	}
+
+	switch k.Spec.Exporter.Method {
+	case "libbpf":
+		configImage += "-libbpf"
+	}
+
 	return &appsv1.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: appsv1.SchemeGroupVersion.String(),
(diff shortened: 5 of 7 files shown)


sthaha:
@BGrasnick , thank you for the PR. @vimalk78 is already working on enabling something similar but using an (undocumented) annotation since we know this (hack) is  going to be short lived. 
The reason I call it a hack is because after 0.7, we will have to deprecate the `method` unless `bcc` is going to be supported for the foreseeable future.

BGrasnick:
Ah ok nevermind then! I wasn't sure if bcc will be deprecated or still be used in the future and thought it might be helpful also in the future because of your answer in #195.
When will this workaround come? We want to roll out kepler but need this or a release of 0.7.0 for kepler to work as we depend on libbpf...

sthaha:
@BGrasnick , it's good to validate our assumptions :) 

@marceloamaral @rootfs, Could you please confirm if `bcc` will be deprecated? In this  PR  @BGrasnick adds a field in exporter spec to allow choosing between bcc and libbpf, depending on which, operator will deploy appropriate image. 

We (@vimalk78 ) are currently working on a solution to allow this through a hack. We intent to add the ability to honour an annotation instead of a field in the spec, so that we don't have to delete the field later on  when bcc won't be used anymore -  which is causes API breakage.

But, this is all made under the assumption that there will only be `libbpf` image in future (from 0.7). Is that the right assumption to make? or would continue to use bcc, or some other mechanism later?

BGrasnick:
@sthaha yeah definitely sounds good to check again what is planned for the future.

My team and organization want to deploy kepler to our cloud/VM-based OpenShift clusters ASAP because we love the integrated dashboards and how easy use to use the operator is. We want to make this available for all teams on our clusters but cannot do it right now because we need to use libbpf instead of bcc. And there is neither a release date for kepler 0.7.0 when libbpf will become default nor a preview for it as far as I know. Do you know of any other way how we could use the libbpf version of kepler with the operator right now?

sthaha:
> My team and organization want to deploy kepler to our cloud/VM-based OpenShift clusters ASAP because we love the integrated dashboards and how easy use to use the operator is. We want to make this available for all teams on our clusters but cannot do it right now because we need to use libbpf instead of bcc. 

@BGrasnick , do note there always exists the possibility of creating your own bundle with *any* kepler image and use operator-sdk to deploy the bundle. 


E.g. here is how I create my bundle that uses kepler libbpf bundle that gets pushed to quay as -  quay.io/sthaha/kepler-operator-bundle:0.0.0-dev

```
make operator-build bundle bundle-build operator-push bundle-push IMG_BASE=quay.io/sthaha VERSION=0.0.0-dev KEPLER_IMG=quay.io/sustainable_computing_io/kepler:release-0.6.1-libbpf
```
There exists a **HACK** as well  (but use it as your own risk)

1. Use community catalog to deploy operator
2. Deploy Kepler
3.  Release all kepler - pod images as follows 
```bash 
for x in $(kubectl get pods -n openshift-kepler-operator -l app.kubernetes.io/component=exporter -l app.kubernetes.io/managed-by=kepler-operator -o name); do
 oc set image -n openshift-kepler-operator $x kepler-exporter=quay.io/sustainable_computing_io/kepler:release-0.6.1-libbpf ;
done
```
We have observed that daemonset controllers do not revert back the image.

marceloamaral:
We might have only `libbpf` in the future, there is no reason, as far as I know, for maintaining duplicated code.
However, I am not sure when we will remove the `bcc`. In the upcoming release, we plan to set `libbpf` as the default, and perhaps in the subsequent release, we may consider removing `bcc`.

Are there any use cases that I might not be aware of that require `bcc`? 
If not, it's probably unnecessary to enable it in the operator.

sthaha:
@marceloamaral , thank you for taking time to answer. 

> Are there any use cases that I might not be aware of that require bcc?

Not that I am aware of but definitely there are more use-cases for `libbpf` than `bcc`. This also makes me wonder if it would be possible for us to make a release of kepler 0.7.0 that only switches to `libbpf` provided libbpf transition is mature enough. 
This would allow us to make another operator release which deploys libbpf based `kepler` allows users of openshift 4.11+ to try out kepler and provide feedback. 


> If not, it's probably unnecessary to enable it in the operator.

@BGrasnick, I agree with Marcelo's comment. Since this issue is short lived, the hack @vimalk78 is working on should suffice.

BGrasnick:
@marceloamaral thanks for your answer!

@sthaha yeah of course that makes sense! If kepler doesn't need/support bcc in the future then there is no meaning in implementing this functionality of method choosing. However, as you outlined it would be great to fast track a release of kepler and kepler-operator with libbpf. I don't know if other people already use kepler on OpenShift + AWS with bcc but for us it does not work...

sthaha:
@BGrasnick I have started a discussion about switching to libbpf - https://github.com/sustainable-computing-io/kepler/discussions/1027

BGrasnick:
@sthaha That's a great idea! Alternatively, if it does not match up with the release plan for kepler, how about releasing a new kepler-operator 0.10.0 or 0.9.1 where we use release-0.6.1-libbpf of kepler?

sthaha:
> how about releasing a new kepler-operator 0.10.0 or 0.9.1 where we use release-0.6.1-libbpf of kepler?

I am happy with that proposal if we can validate in kepler CI that both images produce the same / comparable values for all metrics.

sthaha:
@BGrasnick , I am closing this in favour of https://github.com/sustainable-computing-io/kepler-operator/pull/293

PR #288, opened 2023-10-23
test(ci): update taints and tolerations test

vprashar2929:
This PR updates the test for taints and tolerations to use only nodes that don't have any taints applied.

### DIFF
--- a/pkg/utils/test/framework.go
+++ b/pkg/utils/test/framework.go
@@ -19,12 +19,12 @@ package test
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
-	"strings"
 	"sync"
 	"testing"
 	"time"
+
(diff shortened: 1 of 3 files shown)


pkg/utils/test/framework.go
```diff
@@ -249,26 +232,34 @@ func (f Framework) TaintNode(node, taintStr string) error {
...
+	for _, taint := range taints {
+		if taint.Effect == effect {
+			return true
+		}
+	}
+	return false
 }
 
-func (f Framework) GetTaintsForNode(node string) []corev1.Taint {
+func (f Framework) GetScheduleNodes() []corev1.Node {
```
sthaha, line 245:
```suggestion
func (f Framework) GetSchedulableNodes() []corev1.Node {
```

sthaha (approved):

PR #293, opened 2023-10-26
feat: Use Kepler annotation to choose bcc or libbpf kepler image

vimalk78:
This PR reads value of annotation `"app.kubernetes.io/kepler-bpf-attach-method"` on Kepler object to decide which bpf type image to be deployed.
If the value is `"libbpf"` , the image `quay.io/sustainable_computing_io/kepler:release-0.6.1-libbpf` is chosen, else `quay.io/sustainable_computing_io/kepler:release-0.6.1` is used. 

This is a temporary fix to let users choose libbpf image till kepler upstream makes libbpf as the default attach method.

fixes: https://issues.redhat.com/browse/POWERMON-166

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ GOARCH := $(shell go env GOARCH)
 VERSION ?= $(shell cat VERSION)
 
 KEPLER_VERSION ?=release-0.6.1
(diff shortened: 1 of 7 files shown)


sthaha (approved):

BGrasnick:
This is great, thank you @vimalk78 @sthaha!

PR #294, opened 2023-10-26
Minor corrections and Doc changes to bpf deployment

sthaha:
This PR includes 3 minor corrections / modifications 
1. Fix `make run-k8s` which had an additional `-` in the arg by replacing both `run` and `run-k8s` with just a single `run` target `make run OPENSHIFT=[true]|false`
3. Use kepler.sustainable.computing.io namespace for kepler related annotations 
4. Add a comment in kepler sample explaining how to enable libbpf


vimalk78 (approved):
lgtm

PR #301, opened 2023-11-02
ci: test bundle-upgrade scenario

sthaha:
Previously, in CI, the operator bundle was installed on a clean cluster instead of the scenario where the bundle is upgraded. This has resulted in bundle getting rejected in Community Operator Catalog when the upgrade test failed.

This commit modifies the e2e test to run the upgrade scenario instead of installing on a fresh cluster.

### DIFF
--- a/tests/run-e2e.sh
+++ b/tests/run-e2e.sh
@@ -13,11 +13,14 @@ declare -r OPERATOR="kepler-operator"
 declare -r OLM_CATALOG="kepler-operator-catalog"
 declare -r VERSION="0.0.0-e2e"
 declare -r OPERATOR_DEPLOY_YAML="config/manager/manager.yaml"
+declare -r OPERATOR_CSV="bundle/manifests/$OPERATOR.clusterserviceversion.yaml"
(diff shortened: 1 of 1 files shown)


tests/run-e2e.sh
```diff
@@ -89,19 +92,36 @@ gather_olm() {
...
+		line 50 heavy
 		gather_olm || true
+		fail "Running Released Bundle failed"
+		return $ret
 	}
 
-	return $ret
+	info "Running Upgrade to new bundle"
+	run ./tmp/bin/operator-sdk run bundle-upgrade "$BUNDLE_IMG" \
+		--namespace "$OPERATORS_NS" --use-http || {
```
vprashar2929, line 116: why use `--use-http` ?

sthaha: because localhost:5000 is using http

vprashar2929: oh got it

sthaha: https://github.com/operator-framework/operator-sdk/blob/master/internal/olm/operator/registry/index_image.go#L131

vprashar2929 (approved):

PR #303, opened 2023-11-03
chore: update make fresh

vprashar2929:
This PR fixes the issue when running `make fresh` command. `IMG_BASE` points to `localhost:5001`. This var is also used in `KEPLER_IMG` and `KEPLER_IMG_LIBBPF`. At the time of creating the instance cluster tries to pull the Kepler image from local registry which we don't build and push to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,21 +26,20 @@ VERSION ?= $(shell cat VERSION)
 KEPLER_VERSION ?=release-0.6.1
 KEPLER_VERSION_LIBBPF ?=release-0.6.1-libbpf
 
-# IMG_BASE defines the docker.io namespace and part of the image name for remote images.
-# This variable is used to construct full image tags for bundle and catalog images.
-#
-# For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
-# sustainable_computing_io/kepler-operator-bundle:v$VERSION and sustainable_computing_io/kepler-operator-catalog:$VERSION
+# IMG_BASE and KEPLER_IMG_BASE are set to distinguish between Operator-specific images and Kepler-Specific images.
+# IMG_BASE is used for building and pushing operator related images.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,7 +52,9 @@ fresh: ## default target - sets up a k8s cluster with images ready for deploymen
 		operator-build operator-push \
 		bundle bundle-build \
 		bundle-push \
-		IMG_BASE=localhost:5001 VERSION=0.0.0-dev ;\
+		IMG_BASE=localhost:5001 VERSION=0.0.0-dev \
+		KEPLER_IMG=quay.io/sustainable_computing_io/kepler:$(KEPLER_VERSION) \
+		KEPLER_IMG_LIBBPF=quay.io/sustainable_computing_io/kepler:$(KEPLER_VERSION_LIBBPF)  ;\
```
sthaha, line 57: I think we should fix how `KEPLER_IMAGE`(s) are initialised. How about we create a `KEPLER_IMG_BASE` which always points to `quay.io/...` by default and use `IMG_BASE` only for OPERATOR related images ?

Makefile
```diff
@@ -33,10 +33,11 @@ KEPLER_VERSION_LIBBPF ?=release-0.6.1-libbpf
 # sustainable_computing_io/kepler-operator-bundle:v$VERSION and sustainable_computing_io/kepler-operator-catalog:$VERSION
 IMG_BASE ?= quay.io/sustainable_computing_io
 
+KEPLER_OPERATOR_IMG_BASE :=$(IMG_BASE)
```
sthaha, line 36: sorry, if I wasn't clear last time. We want to `IMG_BASE` for operator image. and `KEPLER_IMG_BASE` for all **kepler** related images. 👼 .

sthaha (approved):
Thank you ❤️

Issue #302, opened 2023-11-02
Make it possible for OpenShift users to view the kepler metrics of their namespace in the developer console


BGrasnick:
If I understand correctly, as of now only administrators can view the Power Monitoring dashboards that are integrated into the OpenShift Console Observe => Dashboards section.
Is it possible to allow users / developers to see the kepler metrics regarding their power consumption in the developer section of the OpenShift Console?
What steps need to be taken to achieve this?

Probably adding some label like `console.openshift.io/odc-dashboard: 'true'` to the grafana-dashboard-k8s-resources-namespace ConfigMap in the openshift-config-managed namespace? And exporting the container namespaces as namespaces? Is this even achievable not to difficultly?

It would be very nice for users of multi tenant OpenShift clusters and could help the productization of kepler inside OpenShift.

BGrasnick:
Any plans or ideas @sthaha @vimalk78 @vprashar2929 @jgbernalp?

sthaha:
@BGrasnick That is indeed a great suggestion! I am not sure how straight forward this is going to be. I am not sure if servicemonitor deployed to `openshift-kepler-operator`  can be used to relabel `container_namespace` to `namespace` 

@simonpasquier do you know if this is possible?

PR #307, opened 2023-11-17
Feat: export correct namespace

BGrasnick:
This PR makes the ServiceMonitor relabel namespace to have the same value as container_namespace instead of "openshift-kepler-operator" applied after scraping so that it is more straight-forward to use the metrics.

Also adds label `"openshift.io/cluster-monitoring":    "true"` so that the metrics get scraped by openshift-monitoring instead of user workload monitoring which makes usage more straight-forward. 
I am not sure if this is wanted for kepler-operator but in our use case it would be much better. In case this is not the intended behaviour, we can just revert the corresponding commit.

Was tested and works on OpenShift 4.11 cluster.

Addresses #302

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -57,6 +57,7 @@ func NewKeplerNamespace() *corev1.Namespace {
 				//   allowPrivilegeEscalation != false (container "kepler-exporter" must set
 				//   securityContext.allowPrivilegeEscalation=false),
 				"pod-security.kubernetes.io/enforce": "privileged",
+				"openshift.io/cluster-monitoring":    "true",
 			}),
 			//TODO: ensure in-cluster monitoring ignores this ns
 		},
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -57,7 +57,6 @@ func NewKeplerNamespace() *corev1.Namespace {
 				//   allowPrivilegeEscalation != false (container "kepler-exporter" must set
 				//   securityContext.allowPrivilegeEscalation=false),
 				"pod-security.kubernetes.io/enforce": "privileged",
-				"openshift.io/cluster-monitoring":    "true",
 			}),
 			//TODO: ensure in-cluster monitoring ignores this ns
 		},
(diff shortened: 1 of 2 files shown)


pkg/components/components.go
```diff
@@ -57,6 +57,7 @@ func NewKeplerNamespace() *corev1.Namespace {
 				//   allowPrivilegeEscalation != false (container "kepler-exporter" must set
 				//   securityContext.allowPrivilegeEscalation=false),
 				"pod-security.kubernetes.io/enforce": "privileged",
+				"openshift.io/cluster-monitoring":    "true",
```
sthaha, line 60: @BGrasnick, we shouldn't use incluster monitoring for operator that are not part of the platform itself. This reserved for only those operators. Rest of the projects / operators must use UWM or some other solution 

cc: @simonpasquier, please correct me if I am wrong 🙏

simonpasquier: Correct, patching the namespace with the `openshift.io/cluster-monitoring` label should happen either in the downstream version of the operator or it should be an option of the upstream operator (disabled by default). It is also something that isn't relevant for non-OCP environments.

https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/#collecting-metrics-with-prometheus

sthaha: Thanks @simonpaquier for the link 🙇

pkg/components/exporter/exporter.go
```diff
@@ -489,6 +489,14 @@ func NewServiceMonitor() *monv1.ServiceMonitor {
...
 	}}
 
+	metricRelabelings := []*monv1.RelabelConfig{{
+		Action: "replace",
+		SourceLabels: []monv1.LabelName{
+			"container_namespace",
+		},
+		TargetLabel: "namespace",
+	}}
+
```
sthaha, line 499: @BGrasnick Have you tested this on OpenShift? IIUC, this will not be honoured in openshift if you use User Project Monitoring.

sthaha:
@BGrasnick I am going to close this due to the reason mentioned above. Please feel free to reopen if you have found a solution for this. Meanwhile we have added the dashboards to developer view as well. So any user with `view` role to the kepler deployment namespace should be able to see all metrics. 
See #369

PR #309, opened 2023-11-21
feat: allow multiple kepler deployments using kepler-internal

sthaha:
**Important Changes:**
* allows setting `deployment-namespace` instead of hardcoded `openshift-kepler-operator`
* modifies the default kepler deployment namespace to `kepler-operator` 
* Allows watching additional namespaces using `watch-namepaces=foo,bar --watch-namespaces=baz` (multiple list)
* Adds `status.exporter` to indicate kepler deployment `conditions` 
* 

** Testing the PR locally** 
To give this a spin on `kind` cluster created using `make fresh` 

```sh
operator-sdk run bundle quay.io/sthaha/kepler-operator-bundle:0.0.0-kepler-internal \
  --namespace=operators \
  --use-http
```


pkg/components/exporter/exporter.go
```diff
@@ -37,40 +36,29 @@ import (
...
+	// FQServiceAccountName = "system:serviceaccount:" + k.Namespace() + ":" + ServiceAccountName
+	//
+	// ClusterRoleName        = prefix + "cr"
+	// ClusterRoleBindingName = prefix + "crb"
 
-	ClusterRoleName        = prefix + "cr"
-	ClusterRoleBindingName = prefix + "crb"
-
-	ConfigmapName = prefix + "cm"
-	DaemonSetName = prefix + "ds"
```
vprashar2929, line 51: Shouldn't we keep `ds` name constant as `kepler-exporter` ?

vprashar2929:
In case of multiple kepler deployments we have to figure out  openshift dashboard's deployments and also revise our prometheus rules accordingly.

pkg/components/exporter/exporter.go
sthaha: what would happen if the 2 keplerinternals are deployed to the the same namespace with the daemonset name set as a const 💭 ?

sthaha:
> In case of multiple kepler deployments we have to figure out openshift dashboard's deployments and also revise our prometheus rules accordingly.

Thats not a supported scenario to start with. I am thinking of adding a
```
spec:
  dashboards: 
      enabled: t|f
  exporter:
     ....
```
to KeplerInternal where `dashboards.enabled` defaults to `false`. I also think the prometheusrule may be giving wrong values since we are not filter out by service.

sthaha:
@vprashar2929 I have added an `openshift` specific api now to enable / disable dashboards.

pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+type DashboardSpec struct {
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+type OpenShiftSpec struct {
+	// +kubebuilder:default=false
+	Enabled   bool          `json:"enabled,omitempty"`
+	Dashboard DashboardSpec `json:"dashboard,omitempty"`
```
vimalk78, line 51: for openshift, can we keep the `Enabled` default value to true?

cmd/manager/main.go
```diff
@@ -59,26 +59,45 @@ func init() {
...
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 
+	flag.StringVar(&controllers.KeplerDeploymentNS, "deployment-namespace", controllers.KeplerDeploymentNS,
+		"Namespace where kepler and its components are deployed.")
+
+	flag.CommandLine.Var(flag.Value(&additionalNamespaces), "watch-namespaces",
```
vimalk78, line 90: can we elaborate a bit how this parameter is to be used? is this for running multiple kepler-exporters?

bundle/manifests/kepler-operator.clusterserviceversion.yaml
```diff
@@ -340,5 +358,5 @@ spec:
     name: kepler
   - image: quay.io/sustainable_computing_io/kepler:release-0.6.1-libbpf
     name: kepler-libbpf
-  replaces: kepler-operator.v0.9.0
+  replaces: kepler-operator.v0.0.0-kepler-internal
```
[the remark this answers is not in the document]
sthaha: Thank you! good catch

cmd/manager/main.go
sthaha: I am also thinking of allowing a `*` as one of the options to allow watching all namespaces but lets do it when we really find a need.

pkg/api/v1alpha1/kepler_internal_types.go
sthaha: Sure ... Internal API can break at any time :D

pkg/components/exporter/exporter.go
```diff
@@ -94,57 +57,62 @@ var (
...
 			},
 		}
 	}
 
-	deployment := k.Spec.Exporter.Deployment
+	deployment := k.Spec.Exporter.Deployment.ExporterDeploymentSpec
+	image := k.Spec.Exporter.Deployment.Image
 	nodeSelector := deployment.NodeSelector
 	tolerations := deployment.Tolerations
+	port := deployment.Port
```
vimalk78, line 80: for multiple kepler deployments, the `port` also has to be different. Users must ensure that.

bundle/manifests/kepler-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
vimalk78: it was easiest to catch  :D

tests/e2e/kepler_test.go
vimalk78, line 1: can we add a similar e2e test for keplerinternal ?

vimalk78: may be in another PR.

vimalk78 (approved):
lgtm

Issue #313, opened 2023-11-24
make cluster-up failed due to DNS


jichenjc:
looks like this is different to https://github.com/sustainable-computing-io/kepler-operator/issues/260
I believe it's due to DNS settings and the worker node is not able to resolve kind-control-plane
can someone help proivde some suggestions?

```
Command Output: I1124 00:07:53.844056     175 join.go:412] [preflight] found NodeName empty; using OS hostname as NodeName
I1124 00:07:53.844127     175 joinconfiguration.go:76] loading configuration from "/kind/kubeadm.conf"
I1124 00:07:53.845320     175 controlplaneprepare.go:225] [download-certs] Skipping certs download
I1124 00:07:53.845351     175 join.go:529] [preflight] Discovering cluster-info
I1124 00:07:53.845679     175 token.go:80] [discovery] Created cluster-info discovery client, requesting info from "kind-control-plane:6443"
I1124 00:07:53.848887     175 round_trippers.go:553] GET https://kind-control-plane:6443/api/v1/namespaces/kube-public/configmaps/cluster-info?timeout=10s  in 1 milliseconds
I1124 00:07:53.848946     175 token.go:217] [discovery] Failed to request cluster-info, will try again: Get "https://kind-control-plane:6443/api/v1/namespaces/kube-public/configmaps/cluster-info?timeout=10s": dial tcp: lookup kind-control-plane on 9.20.136.11:53: no such host

```

jichenjc:
@sthaha any insight on this or suggestions? Thanks

vprashar2929:
@jichenjc Are you using docker or podman? From the logs it looks similar to this issue: https://github.com/kubernetes-sigs/kind/issues/3412

jichenjc:
I am using docker.. is it need to be podman ?

vprashar2929:
No, the issue is with podman. With docker it should work🤔

sthaha:
@jichenjc, does make cluster-up work for you from the `kepler` repo?

jichenjc:
um.. seems not , I got exactly same issue , so definitely something need to be updated 
I always use kind create one k8s node before but with operator I have to follow the `make cluster-up` way..
so first time see this problem

sthaha:
`kind create` should work for operator as well but a lot of things that the `cluster-up` does will have to manually done.

jichenjc:
I am stucking at 2nd kind creation (because the kind is not able to join the first kind ) due to DNS setup
[...]

PR #316, opened 2023-12-01
update kind from 0.12=>0.20

jichenjc:
https://kind.sigs.k8s.io/ latest release is 0.2.0
nothing mandatory but seems reasonable to upgrade

### DIFF
--- a/hack/integration_cluster.sh
+++ b/hack/integration_cluster.sh
@@ -25,7 +25,7 @@ main() {
 
 function kindTest() {
 	echo "install kind"
-	go install sigs.k8s.io/kind@v0.12.0
+	go install sigs.k8s.io/kind@v0.20.0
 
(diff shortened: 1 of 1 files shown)


vprashar2929:
@jichenjc We don't use `hack/integration_cluster.sh` while setting up a cluster for local development. `make cluster-up` makes use of [local-dev-cluster](https://github.com/sustainable-computing-io/local-dev-cluster) for configuring the cluster. The installation of Kind cluster is up to user. Here is the link to the prerequisites: https://github.com/sustainable-computing-io/local-dev-cluster#prerequisites

jichenjc:
ok,  I saw this in another fix (I am trying to fix https://github.com/sustainable-computing-io/kepler-operator/issues/313 and find this might help) , but agree with you on the comments, thanks~

PR #319, opened 2023-12-04
fix: run all tests in CI

sthaha:
This PR includes the following fixed 
* run all tests instead of just the reconciliation test
* provide more time for operator to be ready before running tests
* fix the broken invalid resource name that used wrong namespace for kepler deployment


vprashar2929 (approved):
LGTM

PR #320, opened 2023-12-05
test: improve e2e test script

sthaha:
TODO: summarize this


Makefile
```diff
@@ -265,6 +265,10 @@ CRDOC ?= $(LOCALBIN)/crdoc
 tools:
 	@./hack/tools.sh
 
+.PHONY: kubectl
+kubectl: ## Download kubectl locally if necessary.
+	@./hack/tools.sh $@
```
vprashar2929, line 270: This should be `@./hack/tools.sh kubectl`

hack/tools.sh
```diff
@@ -44,6 +45,60 @@ declare -r OC_URL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/$OC
...
+	kubectl version --client
+}
+
+install_kubectl() {
+	local version_regex="Client Version: $KUBECTL_VERSION"
+
+	validate_version kubectl "version --client" "$version_regex" && return 0
+
+	info "installing kubectl version: $KUBECTL_VERSION"
+	local install_url="https://dl.k8s.io/release/$KUBECTL_VERSION/bin/$GOOS/amd64/kubectl"
```
vprashar2929, line 88: Shouldn't we replace `amd64` with `$GOARCH`?

hack/tools.sh
```diff
@@ -87,13 +142,23 @@ install_operator-sdk() {
 	ok "operator-sdk was installed successfully"
 }
 
+version_govulncheck() {
+	warn govulncheck - latest
```
vprashar2929, line 146: This shouldn't be `warn` right?

hack/tools.sh
```diff
@@ -44,6 +45,60 @@ declare -r OC_URL="https://mirror.openshift.com/pub/openshift-v4/clients/ocp/$OC
...
+	}
+
+	ok "$cmd matching $version_regex already installed"
+}
+
+version_kubectl() {
+	kubectl version --client
+}
+
+install_kubectl() {
```
vprashar2929, line 82: Is there any specific reason why we are installing `kubectl` separately? we get `kubectl` when we install `oc`

tests/run-e2e.sh
```diff
@@ -124,6 +125,22 @@ run_bundle_upgrade() {
...
+run_bundle() {
+	header "Running Bundle"
+	local -i ret=0
+
+	run ./tmp/bin/operator-sdk run bundle "$BUNDLE_IMG" \
+		--namespace "$OPERATORS_NS" --use-http || {
+		ret=$?
+		line 50 heavy
+		gather_olm || true
+		fail "Running Bundle Upgrade failed"
```
vprashar2929, line 137: This should be `"Running Bundle failed"`

Makefile
sthaha: no .. that works

hack/tools.sh
sthaha: Yes, the latest supports waiting for `status.exporter` to be set.

hack/tools.sh
sthaha: Fixed

vprashar2929 (approved):
LGTM

PR #235, opened 2023-09-14
update model server support

sunya-ch:
This PR updates model server support aiming for release v0.6 as mentioned in https://github.com/sustainable-computing-io/kepler-operator/issues/232. 

API doc: 
https://github.com/sustainable-computing-io/kepler-operator/blob/c15c77621958cc79d1921d9af378915158abc4ca/docs/api.md

The PR contains changes on :
- updating KeplerSpec
- updating modelserver component
- adding estimator component
- modifying exporter to have sidecar if configure
- adding modelServerReconcilers

Note that, 
The holder for setting filters and model name is here on kepler: 
https://github.com/sustainable-computing-io/kepler/blob/73cb11fb963f425013cf7f03f214c8f8b85c7853/pkg/config/config.go#L390.
However, it is not determined how to use it. So, it is not supported yet from end to end.

Example configmap change from full deployment on OpenShift on IBM Cloud 
(kepler CR: `config/samples/kepler_full_deploy.yaml`)

```yaml
> oc get configmap -n openshift-kepler-operator model-server-cm -oyaml
...
  MODEL_CONFIG: |
    NODE_COMPONENTS_ESTIMATOR=true
    NODE_COMPONENTS_INIT_URL=https://raw.githubusercontent.com/sustainable-computing-io/kepler-model-db/main/models/Linux-4.15.0-213-generic-x86_64_v0.6/rapl/AbsPower/KubeletOnly/GradientBoostingRegressorTrainer_1.zip
  MODEL_SERVER_ENABLE: "true"
  MODEL_SERVER_URL: http://model-server-svc.openshift-kepler-operator.svc.cluster.local:8100
```

Resources:

```bash
> oc get -n openshift-kepler-operator all
NAME                                       READY   STATUS    RESTARTS   AGE
pod/kepler-exporter-ds-48lff               2/2     Running   0          4m10s
pod/kepler-exporter-ds-4nj8g               2/2     Running   0          4m10s
pod/kepler-exporter-ds-5lj62               2/2     Running   0          4m10s
pod/kepler-exporter-ds-9rqlv               2/2     Running   0          4m10s
pod/kepler-exporter-ds-knnsh               2/2     Running   0          4m10s
pod/kepler-exporter-ds-kskmq               2/2     Running   0          4m11s
pod/kepler-exporter-ds-szl8w               2/2     Running   0          2m49s
pod/model-server-deploy-85fd7b8c6d-wkkrg   1/1     Running   0          4m10s

[...]

Issue #312, opened 2023-11-23
Add config map for controller manager


vprashar2929:
With the addition of Kepler Internals it make sense to add a config map so that any dynamic changes are respected by controller manager.

PR #322, opened 2023-12-07
add model-server/estimator to KeplerInternal

sunya-ch:
This PR replaces https://github.com/sustainable-computing-io/kepler-operator/pull/235 by moving the integration to Kepler-internal API. 

**Change summary:**
- Add Estimator and ModelServer in `KeplerInternalSpec` and `KeplerInternalStatus`
- Add components/estimator
- Implement components/model-server when enabled
- Add Model Server Reconcilers to kepler-internal Reconciler
- Modify components/exporter to add estimator sidecar if set 
- Add access Role for deployments and persistentvolumeclaims
- Add `AddIfNotEmpty` and `VolumeFromEmptyDir` utility function (used for Estimator and ModelServer creation)

**Bug fixes (not related to model server):**
- Correct reference for `UP-TO-DATE` status
- Replace ki.namespace in updateStatus with the deployment namespace from spec.

---
## KeplerInternal
Here is the CR that I used for running in my local cluster.
```yaml
apiVersion: kepler.system.sustainable.computing.io/v1alpha1
kind: KeplerInternal
metadata:
  annotations:
    kepler.sustainable.computing.io/bpf-attach-method: libbpf
  labels:
    app.kubernetes.io/name: kepler
    app.kubernetes.io/instance: kepler
    app.kubernetes.io/part-of: kepler-operator
  name: kepler
spec:
  exporter:
    deployment:
      image: quay.io/sustainable_computing_io/kepler:release-0.6.1-libbpf
      namespace: kepler-operator
  openshift:
    enabled: true
    dashboard:
      enabled: true
  modelServer:
    enabled: true
  estimator:
    node:
      components:
        sidecar: true
        initUrl: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip
      total:
        sidecar: true
        initUrl: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip
```

## KeplerInternal Status

With neither estimator nor modelserver
```
NAME     PORT   DESIRED   CURRENT   UP-TO-DATE    READY   AVAILABLE   AGE   IMAGE   ESTIMATOR      MODEL-SERVER
kepler   9103   7         7         7             7       7           83s   <abbr>  NotInstalled   NotInstalled
```

With only estimator
```
NAME     PORT   DESIRED   CURRENT   UP-TO-DATE    READY   AVAILABLE   AGE   IMAGE   ESTIMATOR      MODEL-SERVER
kepler   9103   7         7         7             7       7           40m   <abbr>  Running        NotInstalled
```

With estimator+modelserver
```
NAME     PORT   DESIRED   CURRENT   UP-TO-DATE     READY   AVAILABLE   AGE   IMAGE    ESTIMATOR     MODEL-SERVER
kepler   9103   7         7         7              7       7           15s   <abbr>   Running        Running
```

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>


pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -53,8 +54,71 @@ type OpenShiftSpec struct {
...
+	Exporter    InternalExporterSpec     `json:"exporter"`
+	Estimator   *InternalEstimatorSpec   `json:"estimator,omitempty"`
+	ModelServer *InternalModelServerSpec `json:"modelServer,omitempty"`
+	OpenShift   OpenShiftSpec            `json:"openshift,omitempty"`
+}
+
+// Kepler Model Server Spec
+type InternalModelServerSpec struct {
+
+	// +kubebuilder:default=true
```
sthaha, line 66: could you please keep the default to `false` for all features that are not part of stable api - `kepler` ?

pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -53,8 +54,71 @@ type OpenShiftSpec struct {
...
+	Enabled bool `json:"enabled,omitempty"`
+
+	Image string `json:"image,omitempty"`
+
+	// +kubebuilder:default=""
+	URL string `json:"url,omitempty"`
+
+	// +kubebuilder:default=8100
```
sthaha, line 74: could you please add validation on min and max ?

pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -53,8 +54,71 @@ type OpenShiftSpec struct {
...
+	RequestPath string `json:"requestPath,omitempty"`
+
+	// +kubebuilder:default=""
+	ListPath string `json:"listPath,omitempty"`
+
+	// +kubebuilder:default=""
+	PipelineURL string `json:"pipelineUrl,omitempty"`
+
+	// +kubebuilder:default=""
+	ErrorKey string `json:"errKey,omitempty"`
```
sthaha, line 90: Is this used?

pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -64,13 +128,15 @@ type KeplerInternalSpec struct {
 // +kubebuilder:printcolumn:name="Port",type=integer,JSONPath=`.spec.exporter.deployment.port`
 // +kubebuilder:printcolumn:name="Desired",type=integer,JSONPath=`.status.exporter.desiredNumberScheduled`
 // +kubebuilder:printcolumn:name="Current",type=integer,JSONPath=`.status.exporter.currentNumberScheduled`
-// +kubebuilder:printcolumn:name="Up-to-date",type=integer,JSONPath=`.status.updatedNumberScheduled`
+// +kubebuilder:printcolumn:name="Up-to-date",type=integer,JSONPath=`.status.exporter.updatedNumberScheduled`
```
sthaha, line 131: Thanks :)

pkg/components/exporter/exporter.go
```diff
@@ -249,6 +218,40 @@ func NewConfigMap(d components.Detail, k *v1alpha1.KeplerInternal) *corev1.Confi
...
+		"REDFISH_SKIP_SSL_VERIFY":           "true",
+		"MODEL_CONFIG":                      modelConfig,
+	}
+
+	ms := k.Spec.ModelServer
+	if ms != nil {
+		if ms.Enabled {
+			exporterConfigMap["MODEL_SERVER_ENABLE"] = "true"
+		}
+		modelServerConfig := modelserver.ModelServerConfigForClient(k.ModelServerDeploymentName(), k.Spec.ModelServer)
```
sthaha, line 251:
```suggestion
		modelServerConfig := modelserver.ConfigForClient(k.ModelServerDeploymentName(), k.Spec.ModelServer)
```
How about renaming `modelserver.ModelServer<X>` to `modelserver.<X>` to avoid stuttering ?

sthaha: I wonder if `k.Spec.ModelServer.ClientConfig()` is  better since all the function only requires the spec from  Modelserver to compute client config.

sthaha:
@sunya-ch does this support deployment of multiple model-servers like it currently allows multiple `kepler-internals` 

May be worth considering is ..  if it makes sense to have a separate CRD for KeplerModelServer. This would allow for multiple model servers to be deployed (if that is even a case) with different configs. 
`kepler-internal` can have a `spec.modelserver.ref` to refer to the `KeplerModelServer`

pkg/api/v1alpha1/kepler_internal_types.go
sunya-ch: Done. Thanks.

pkg/api/v1alpha1/kepler_internal_types.go
sunya-ch: Done. I copied the exporter range.

pkg/components/exporter/exporter.go
sunya-ch: Yes..it is the config for client of the model server (i.e., exporter/estimator container).

sunya-ch:
> @sunya-ch does this support deployment of multiple model-servers like it currently allows multiple `kepler-internals`
> 
> May be worth considering is .. if it makes sense to have a separate CRD for KeplerModelServer. This would allow for multiple model servers to be deployed (if that is even a case) with different configs. `kepler-internal` can have a `spec.modelserver.ref` to refer to the `KeplerModelServer`

Yes, I generate model server name in the same way of the kepler exporter name (based on CR + suffix of `model-server`). I think we can keep it together for simplicity of deployment. Each export can connect to different model server. The model server will be created only when it is enabled. But, kepler can specify only model server URL to connect to the other model server (including external).

pkg/controllers/kepler_internal.go
```diff
@@ -199,7 +200,7 @@ func (r KeplerInternalReconciler) updateReconciledStatus(ctx context.Context, ki
 func (r KeplerInternalReconciler) updateAvailableStatus(ctx context.Context, ki *v1alpha1.KeplerInternal, recErr error) {
 	// get daemonset owned by kepler
 	dset := appsv1.DaemonSet{}
-	key := types.NamespacedName{Name: ki.DaemonsetName(), Namespace: ki.Namespace()}
+	key := types.NamespacedName{Name: ki.DaemonsetName(), Namespace: ki.Spec.Exporter.Deployment.Namespace}
```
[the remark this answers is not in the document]
sthaha: @sunya-ch , How about we move `spec.exporter.deployment.namespace` to `spec.namespace` so that all components of `ki` will use this common field for deployment and can't be empty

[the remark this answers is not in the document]
sunya-ch: First, change ki.Namespace to ki.Spec.Exporter.Deployment.Namespace is another bug I fix.
ki (keplerinternal CR) is cluster-wide. ki.Namespace will give an empty "" namespace. I don't know when this bug occur but it made the status failed to update.

Second, I do agree to move the namespace from exporter.deployment.namespace to outside. But, in that case, we actually show move node-selector and toleration to the same level since it shares between estimator and exporter.  I also notice that it is missing in the internal CR. 
We can keep as is and have the estimator and model server uses it as a reference. Or let the model server has its own namespace. WDYT?

sunya-ch:
> @sunya-ch Overall the code looks good to me one missing part is the e2e test.
> Could you please also add a simple e2e test for validating deployment of model server?

I cannot see the e2e test for the KeplerInternal CR. I think it would become too big change on this PR to add the e2e test for the keplerinternal. Could that be done by the other PR then I could help add the model server part? Actually, I think we can just check the status of KeplerInternal on `.status.modelServer.status` and `.status.estimator.status`.

pkg/controllers/kepler_internal.go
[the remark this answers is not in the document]
sthaha: `ki.Namespace()` must never return an empty string since kubernetes API validation should prevent .t
could you please revert this change?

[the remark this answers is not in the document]
sthaha: > Second, I do agree to move the namespace from exporter.deployment.namespace to outside. But, in that case, we actually show move node-selector and toleration to the same level since it shares between estimator and exporter. 

A seperate Node selector for kepler makes-sense but not for model-server. The use-case for kepler is that it deploys a daemonset and pods are deployed per node. so giving the ability to decide which nodes to monitor is a must have and should be separate to deploying a single instance of model-server in *any* node, shouldn't it?

sthaha:
@sunya-ch 

> I cannot see the e2e test for the KeplerInternal CR. I think it would become too big change on this PR to add the e2e test for the keplerinternal. Could that be done by the other PR then I could help add the model server part?

The current Kepler tests cover almost all the usecase currently supported by kepler-internal, so having a set of tests that replicate what creation of  kepler already does gives us only a low ROI. 

IMHO, all features should be be accompanied by tests that validate most common usecases. It shouldn't be too hard to add an e2e by making a copy of the existing kepler-e2e.

sthaha:
@sunya-ch ,

```
 annotations:
    kepler.sustainable.computing.io/bpf-attach-method: libbpf
```
that isn't required for `kepler-internal` it is only a hack enabled for `kepler` so that stable API users have the ability to deploy libbpf image which are kernel agnostic.

sunya-ch:
> @sunya-ch
> 
> > I cannot see the e2e test for the KeplerInternal CR. I think it would become too big change on this PR to add the e2e test for the keplerinternal. Could that be done by the other PR then I could help add the model server part?
> 
> The current Kepler tests cover almost all the usecase currently supported by kepler-internal, so having a set of tests that replicate what creation of kepler already does gives us only a low ROI.
> 
> IMHO, all features should be be accompanied by tests that validate most common usecases. It shouldn't be too hard to add an e2e by making a copy of the existing kepler-e2e.

Yes, it might be but I still think it should be on another PR for better track. I can rebase this PR from the PR.

pkg/controllers/kepler_internal.go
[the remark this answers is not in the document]
sunya-ch: Yes, model server is separated but estimator shared the same. What I think is maybe we can keep the .deployment in exporter spec with namespace, selector, toleration. 
Meanwhile, model-sever may also have its own deployment spec which have a namespace.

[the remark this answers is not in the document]
sunya-ch: > `ki.Namespace()` must never return an empty string since kubernetes API validation should prevent .t
> could you please revert this change? If you experience an issue with it, lets fix that separately.

I just noticed it is the function call. I think I mixed the issue. Now revert back.

sthaha:
@sunya-ch Hopefull https://github.com/sustainable-computing-io/kepler-operator/pull/325 should help with the model-server testing 👼

pkg/components/exporter/exporter.go
sunya-ch: @sthaha Please confirm whether the change resolved?

sunya-ch:
Sorry for multiple force-pushed. 
Most of the main code is not changed the failure comes from my bad code on test case.

Additional change are adding e2e test case and Enabled() function for InternalEstimatorSpec.

https://github.com/sustainable-computing-io/kepler-operator/blob/a4177f9db47817b9a78028e129d37fbbdeb65dfa/pkg/api/v1alpha1/kepler_internal_types.go#L108

tests/e2e/kepler_internal_test.go
```diff
@@ -54,19 +73,114 @@ func TestKeplerInternal_Reconciliation(t *testing.T) {
 	assert.Equal(t, 1, len(containers))
 	assert.Equal(t, 1, len(containers[0].Ports))
 	assert.EqualValues(t, 9108, containers[0].Ports[0].ContainerPort)
+	// test expected status
+	testKeplerInternalStatus(t, f, ki)
```
sthaha, line 77: @sunya-ch , could you please add the assertions to framework instead? `f.AssertInternalStatus`

tests/e2e/kepler_internal_test.go
```diff
@@ -54,19 +73,114 @@ func TestKeplerInternal_Reconciliation(t *testing.T) {
...
+}
 
-	// the status will be updated
-	ki = f.WaitUntilInternalCondition(ki.Name, v1alpha1.Reconciled, v1alpha1.ConditionTrue)
-	assert.Equal(t, []corev1.Toleration{{Operator: "Exists"}}, ki.Spec.Exporter.Deployment.Tolerations)
+func TestKeplerInternal_WithEstimator(t *testing.T) {
+	f := test.NewFramework(t)
+	name := "e2e-kepler-internal-est"
```
sthaha, line 82:
```suggestion
	name := "e2e-ki-modelserver-test"
```

sunya-ch: Why we need the suffix "-test"? If you think we should make kepler-internal in short, how's about `e2e-ki-estimator`, `e2e-ki-modelserver`, `e2e-ki-estimator-modelserver` for the newly-introduced case? (btw, it will have a different format from the basic case which is `e2e-kepler-internal`).

tests/e2e/kepler_internal_test.go
sunya-ch: Move to the assertions.go: https://github.com/sustainable-computing-io/kepler-operator/compare/e75fcca04c20ae4911abdedbedd383b6c6bd76aa..3238a90270873cf411818cae59f0b41e505b74c3

pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -53,8 +54,81 @@ type OpenShiftSpec struct {
...
+	return e.Node.Enabled() || e.Container.Enabled()
+}
+
+type EstimatorGroup struct {
+	Total      *EstimatorConfig `json:"total,omitempty"`
+	Components *EstimatorConfig `json:"components,omitempty"`
+}
+
+func (g EstimatorGroup) Enabled() bool {
+	return g.Total.SidecarEnabled || g.Components.SidecarEnabled
```
sthaha, line 118: This can panic if `Total` is nil. Same for `Components`

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/sustainable.computing.io/kepler-operator/pkg/api/v1alpha1"
+	"github.com/sustainable.computing.io/kepler-operator/pkg/utils/k8s"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	prefix = "estimator-"
+	// exported
+	PVCName = prefix + "pvc"
```
sthaha, line 32: Is this getting used?

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+		MountPath: "/tmp",
+	})
+}
+
+func addSocketWaitCmd(exporterContainer *corev1.Container) *corev1.Container {
+	cmd := exporterContainer.Command
+	exporterContainer.Command = bashCommand
+	exporterContainer.Args = []string{fmt.Sprintf(waitForSocketCommand, strings.Join(cmd, " "))}
+	return exporterContainer
+}
```
sthaha, line 105: Can we add a Unit Test for this if there isn't one already that tests that the original command is preserved?

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+	prefix = "estimator-"
+	// exported
+	PVCName = prefix + "pvc"
+
+	StableImage          = "quay.io/sustainable_computing_io/kepler_model_server:v0.6"
+	waitForSocketCommand = "until [ -e /tmp/estimator.sock ]; do sleep 1; done && %s"
+)
+
+var (
+	bashCommand = []string{"/bin/sh", "-c"}
```
sthaha, line 39:
```suggestion
	shellCommand = []string{"/bin/sh", "-c"}
```

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// NeedsEstimatorSidecar returns true if any of estimator config has sidecar enabled
+func NeedsEstimatorSidecar(es *v1alpha1.InternalEstimatorSpec) bool {
+
```
sthaha, line 44: Should we add ... ? 

```suggestion
if es == nil {
  return false
}
```

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+		Name:      "mnt",
+		MountPath: "/mnt",
+	}, {
+		Name:      "tmp",
+		MountPath: "/tmp",
+	}}
+
+	return corev1.Container{
+		Image:           image,
+		ImagePullPolicy: corev1.PullAlways,
```
sthaha, line 77: Should this be PullAlways? How about `PullIfNotPresent` ?

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+		Name:            "estimator",
+		VolumeMounts:    mounts,
+		Command:         []string{"python3.8"},
+		Args:            []string{"-u", "src/estimate/estimator.py"},
+	}
+}
+
+// NewContainer returns sidecar additional volumes
```
sthaha, line 85:
```suggestion
// NewVolumes returns sidecar additional volumes
```

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+func NewVolumes() []corev1.Volume {
```
sthaha, line 86:
```suggestion
func Volumes() []corev1.Volume {
```

pkg/components/estimator/estimator.go
```diff
@@ -0,0 +1,138 @@
...
+func estimatorConfig(prefix string, spec v1alpha1.EstimatorConfig) string {
+	config := fmt.Sprintf("%s_ESTIMATOR=%v\n", prefix, spec.SidecarEnabled)
+	if spec.InitUrl != "" {
+		config += fmt.Sprintf("%s_INIT_URL=%s\n", prefix, spec.InitUrl)
+	}
+	return config
+}
+
+func ModelConfig(es *v1alpha1.InternalEstimatorSpec) string {
+	config := ""
```
sthaha, line 122: nit: strings.Builder is better suited for this.

pkg/components/estimator/estimator_test.go
```diff
@@ -0,0 +1,141 @@
+package estimator
```
sthaha, line 1: missing header :(

pkg/components/exporter/exporter.go
```diff
@@ -74,16 +76,33 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
+	exporterContainer := newExporterContainer(k.Name, k.DaemonsetName(), k.Spec.Exporter.Deployment)
+
+	var containers []corev1.Container
+
+	var volumes = []corev1.Volume{
+		k8s.VolumeFromHost("lib-modules", "/lib/modules"),
+		k8s.VolumeFromHost("tracing", "/sys"),
+		k8s.VolumeFromHost("proc", "/proc"),
+		k8s.VolumeFromHost("kernel-src", "/usr/src/kernels"),
+		k8s.VolumeFromHost("kernel-debug", "/sys/kernel/debug"),
```
sthaha, line 95: can we remove this since `/sys` is already mounted as `tracing` ?

pkg/components/exporter/exporter.go
```diff
@@ -74,16 +76,33 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
-	port := deployment.Port
 	// NOTE: since 2 or more KeplerInternals can be deployed to the same namespace,
 	// we need to make sure that the pod selector of each of the DaemonSet
 	// create of each kepler is unique. Thus the daemonset name is added as
 	// label to the pod
 
-	bindAddress := "0.0.0.0:" + strconv.Itoa(int(port))
+	exporterContainer := newExporterContainer(k.Name, k.DaemonsetName(), k.Spec.Exporter.Deployment)
+
+	var containers []corev1.Container
```
sthaha, line 88:
```suggestion
```

pkg/components/exporter/exporter.go
```diff
@@ -74,16 +76,33 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
+	var volumes = []corev1.Volume{
+		k8s.VolumeFromHost("lib-modules", "/lib/modules"),
+		k8s.VolumeFromHost("tracing", "/sys"),
+		k8s.VolumeFromHost("proc", "/proc"),
+		k8s.VolumeFromHost("kernel-src", "/usr/src/kernels"),
+		k8s.VolumeFromHost("kernel-debug", "/sys/kernel/debug"),
+		k8s.VolumeFromConfigMap("cfm", k.Name),
+	} // exporter default Volumes
+
+	containers = []corev1.Container{*exporterContainer}
```
sthaha, line 99:
```suggestion
	exporterContainer := newExporterContainer(...) }
	containers := []corev1.Container{ exporterContainer }
```

pkg/components/exporter/exporter.go
```diff
@@ -74,16 +76,33 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
 	nodeSelector := deployment.NodeSelector
 	tolerations := deployment.Tolerations
-	port := deployment.Port
 	// NOTE: since 2 or more KeplerInternals can be deployed to the same namespace,
 	// we need to make sure that the pod selector of each of the DaemonSet
 	// create of each kepler is unique. Thus the daemonset name is added as
 	// label to the pod
 
-	bindAddress := "0.0.0.0:" + strconv.Itoa(int(port))
+	exporterContainer := newExporterContainer(k.Name, k.DaemonsetName(), k.Spec.Exporter.Deployment)
```
sthaha, line 86: How about `newExporterContainer` return an object instead of a pointer since `nil` is never returned?

pkg/components/exporter/exporter.go
```diff
@@ -74,16 +76,33 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
+
+	var containers []corev1.Container
+
+	var volumes = []corev1.Volume{
```
sthaha, line 90: Aren't these volumes already added by `NewExporterContainer` ?

pkg/components/exporter/exporter.go
sthaha, line 95: we need a better way to handling volumes. - In another PR

 E.g. each `New<X>Container` can return `[]NamedMount`. 

```
type NamedMount string

const (
 HostLibModulesMount NamedMount = "host-lib-modules"
 HostProc  = "host-proc"
 KeplerConfigMapMount = "cm-kepler"
)

func (m HostMount) Volume() corev1.Volume {
   mounts := map[HostMount]string{
       LibModulesMount:  "/lib/modules", 
...
    }
    if strings.StartsWith("host-", m) {
     return k8s.VolumeFromHost(m, mounts[m])
   } else if strings.StartsWith("cm-", m)
     return k8s.VolumeFromConfigMap(m, mounts[m])
  }
 
}
```

pkg/components/modelserver/modelserver_test.go
```diff
@@ -0,0 +1,128 @@
+package modelserver
```
sthaha, line 1: header  / copyright info

pkg/reconciler/runner.go
```diff
@@ -33,6 +33,7 @@ type Runner struct {
 	Logger      logr.Logger
 }
 
+// TODO: make sure that model server container (deployment) is ready before creating kepler daemonset
```
sthaha, line 36: why is this needed? Shouldn't kepler retry for some time?

pkg/utils/test/assertions.go
```diff
@@ -108,3 +110,22 @@ func (f Framework) AssertNoResourceExists(name, ns string, obj client.Object, fn
 		f.T.Errorf("%s (%v) exists after %v", k8s.GVKName(obj), key, opt.WaitTimeout)
 	}
 }
+
+func (f Framework) AssertInternalStatus(ki *v1alpha1.KeplerInternal) {
```
sthaha, line 114:
```suggestion
func (f Framework) AssertInternalStatus(name string) {
```

pkg/utils/test/assertions.go
```diff
@@ -108,3 +110,22 @@ func (f Framework) AssertNoResourceExists(name, ns string, obj client.Object, fn
...
+	// the status will be updated
+	ki = f.WaitUntilInternalCondition(ki.Name, v1alpha1.Reconciled, v1alpha1.ConditionTrue)
```
sthaha, line 116:
```suggestion
	ki = f.WaitUntilInternalCondition(name, v1alpha1.Reconciled, v1alpha1.ConditionTrue)
```

sthaha (changes requested):

sthaha:
@vprashar2929 could you please help validate this on OpenShift ?.. just ensuring the supported usecase of  creating a `kepler` works is good enough.

pkg/components/exporter/exporter.go
sunya-ch: No, there are only VolumeMounts come with container.

pkg/reconciler/runner.go
sunya-ch: Current code kepler does not retry if not found model server. All is done at configuration phase.

pkg/components/estimator/estimator.go
sunya-ch: It should be already covered for estimator sidecar added. Any command and volume should be preserved. 
In the test case, we assume

```
        // assume original commands (simplified version)
	exporterContainer := &corev1.Container{
		Command: []string{"kepler", "-v=1"},
	}
	// assume original volumes (simplified version)
	keplerVolumes := []corev1.Volume{k8s.VolumeFromEmptyDir("kepler-volume")}
```

pkg/components/estimator/estimator.go
sunya-ch: updated.

pkg/components/estimator/estimator.go
sunya-ch: Changed.

pkg/components/estimator/estimator.go
sunya-ch: Changed.

pkg/components/estimator/estimator.go
sunya-ch: Sorry. It is previous code. Removed.

vprashar2929:
Couple of observations while testing this on OpenShift(4.13):

* When deploying KeplerInternals on OpenShift, there is still an issue with showing the appropriate status of Kepler
```
oc get keplerinternals.kepler.system.sustainable.computing.io                                                                               
NAME           PORT   DESIRED   CURRENT   UP-TO-DATE   READY   AVAILABLE   AGE    IMAGE                                  ESTIMATOR   MODEL-SERVER
mykepler-101   9103   0         0                      0                   168m   quay.io/rh_ee_vprashar/kepler:latest
```
* When the estimator is deployed along with the model server it checks for the wrong model-server service name: `mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local` `openshift-kepler-operator` ns doesn't exist
```
set NODE_TOTAL_ESTIMATOR to true.
set NODE_TOTAL_INIT_URL to https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip.
set NODE_COMPONENTS_ESTIMATOR to true.
set NODE_COMPONENTS_INIT_URL to https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip.
clean socket
cannot make request to http://mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local:8100/model: HTTPConnectionPool(host='mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local', port=8100): Max retries exceeded with url: /model (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x7fca34a02430>: Failed to resolve 'mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local' ([Errno -2] Name or service not known)"))
get archived model
get init url https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip
try getting archieved model from URL: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip for AbsPower
<Response [200]>
load model from config:  /mnt/download/acpi/AbsPower
cannot make request to http://mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local:8100/model: HTTPConnectionPool(host='mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local', port=8100): Max retries exceeded with url: /model (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x7fca33e17460>: Failed to resolve 'mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local' ([Errno -2] Name or service not known)"))
get archived model
get init url https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip
try getting archieved model from URL: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip for AbsPower
<Response [200]>
load model from config:  /mnt/download/rapl/AbsPower
```

sunya-ch:
> Couple of observations while testing this on OpenShift(4.13):
> 
> * When deploying KeplerInternals on OpenShift, there is still an issue with showing the appropriate status of Kepler
> 
> ```
> oc get keplerinternals.kepler.system.sustainable.computing.io                                                                               
> NAME           PORT   DESIRED   CURRENT   UP-TO-DATE   READY   AVAILABLE   AGE    IMAGE                                  ESTIMATOR   MODEL-SERVER
> mykepler-101   9103   0         0                      0                   168m   quay.io/rh_ee_vprashar/kepler:latest
> ```
> 
> * When the estimator is deployed along with the model server it checks for the wrong model-server service name: `mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local` `openshift-kepler-operator` ns doesn't exist
> 
> ```
> set NODE_TOTAL_ESTIMATOR to true.
> set NODE_TOTAL_INIT_URL to https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip.
> set NODE_COMPONENTS_ESTIMATOR to true.
> set NODE_COMPONENTS_INIT_URL to https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip.
> clean socket
> cannot make request to http://mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local:8100/model: HTTPConnectionPool(host='mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local', port=8100): Max retries exceeded with url: /model (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x7fca34a02430>: Failed to resolve 'mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local' ([Errno -2] Name or service not known)"))
> get archived model
> get init url https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip
> try getting archieved model from URL: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/acpi/AbsPower/BPFOnly/GradientBoostingRegressorTrainer_1.zip for AbsPower
> <Response [200]>
> load model from config:  /mnt/download/acpi/AbsPower
> cannot make request to http://mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local:8100/model: HTTPConnectionPool(host='mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local', port=8100): Max retries exceeded with url: /model (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x7fca33e17460>: Failed to resolve 'mykepler-101-model-server-svc.openshift-kepler-operator.svc.cluster.local' ([Errno -2] Name or service not known)"))
> get archived model
> get init url https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip
> try getting archieved model from URL: https://raw.githubusercontent.com/sunya-ch/kepler-model-db/css-117/models/v0.6/css-117/core96_v0.6/rapl/AbsPower/BPFOnly/KNeighborsRegressorTrainer_1.zip for AbsPower
> <Response [200]>
> load model from config:  /mnt/download/rapl/AbsPower
> ```

Thank you. I did see the manual namespace set for model server default URL. Now, I updated it to use the namespace from the kepler spec.

For status not updated, could you share describe result? I cannot find the cause of issue since I can see the updated status on my OpenShift cluster (4.12).

pkg/reconciler/runner.go
sthaha: In a different PR, we can write a reconciler that waits for a resource to be available. If modelserver is enabled, inject that before kepler daemonset is created.

pkg/utils/test/framework.go
```diff
@@ -250,6 +250,37 @@ func (f Framework) WaitUntilInternalCondition(name string, t v1alpha1.ConditionT
...
+func (f Framework) WaitUntilInternalHasExpectedRunning(name string) *v1alpha1.KeplerInternal {
+	f.T.Helper()
+	k := v1alpha1.KeplerInternal{}
+	f.WaitUntil(fmt.Sprintf("kepler-internal %s has expected running status", name),
+		func() (bool, error) {
+			err := f.client.Get(context.TODO(), client.ObjectKey{Name: name}, &k)
+			if errors.IsNotFound(err) {
+				return true, fmt.Errorf("kepler-internal %s is not found", name)
+			}
+			statusOK := true
```
sthaha, line 262:
```suggestion
			var statusOK bool
```

vprashar2929:
> Thank you. I did see the manual namespace set for model server default URL. Now, I updated it to use the namespace from the kepler spec.

Should we use the namespace from `keplerinternal` spec ?

> For status not updated, could you share describe result? I cannot find the cause of issue since I can see the updated status on my OpenShift cluster (4.12). From the above result, it seems the issue not only from model server but also the status of the deamonset.

So when I created keplerinternals I used a different namespace from `kepler-operator` AFAIK controller only watches `kepler-operator` ns. If you create an instance with the different namespace to `kepler-operator` then you won't be able to see the status. This is a known issue and we are planning to fix it by adding config-map #312

sthaha:
@sunya-ch 

I see the following difference with the install of kepler using `kepler` CRD from `v1alpha` branch vs this PR. Ideally we require  that optional kepler-internal features do not introduce any difference.

However, I am okay with having this if you could confirm that there is no impact or fixes an issue

### configmap

image

sunya-ch:
> @sunya-ch
> 
> I see the following difference with the install of kepler using `kepler` CRD from `v1alpha` branch vs this PR. Ideally we require that optional kepler-internal features do not introduce any difference.
> 
> However, I am okay with having this if you could confirm that there is no impact or fixes an issue
> 
> ### configmap
> image

Yes, it has no affect because the default is false. Only enable when it is set to true. (reference: https://github.com/sustainable-computing-io/kepler/blob/442bcfe5d3bf26a1285ff0f13d1f5017d28b9e37/pkg/model/model.go#L189)

sunya-ch:
> > Thank you. I did see the manual namespace set for model server default URL. Now, I updated it to use the namespace from the kepler spec.
> 
> Should we use the namespace from `keplerinternal` spec ?
> 
> > For status not updated, could you share describe result? I cannot find the cause of issue since I can see the updated status on my OpenShift cluster (4.12). From the above result, it seems the issue not only from model server but also the status of the deamonset.
> 
> So when I created keplerinternals I used a different namespace from `kepler-operator` AFAIK controller only watches `kepler-operator` ns. If you create an instance with the different namespace to `kepler-operator` then you won't be able to see the status. This is a known issue and we are planning to fix it by adding config-map #312

Finally, I guess so, we may allow Kepler to deploy on any namespace for keplerinternal. 
I put TODO in this PR: https://github.com/sustainable-computing-io/kepler-operator/pull/324/files#diff-435eecb1d2af40ee96747f88ab71eb2758da989c92f76dcb1f714fc1b300c633

We need to have Kepler-operator add the new namespace to the cache. Otherwise, we have to rely on the additional namespace list in the command line. Kepler deployer has to know in advance the namespace list where they are going to allow the keplerinternal to be installed. 

https://github.com/sustainable-computing-io/kepler-operator/blob/73505af1859d9f3f097307a2e035d2f2f33e7ff1/cmd/manager/main.go#L90

sthaha (approved):
Looks great! Thanks a lot @sunya-ch 🤗

PR #329, opened 2023-12-15
fix(kustomize): remove auto-generated clusterPermissions from base CSV

sthaha:
The rbac required by the operators is auto-generated from the `kubebuilder` tags such as below

```
See: pkg/controllers/kepler*.go

//+kubebuilder:rbac:groups=kepler.system.sustainable.computing.io,resources=*,verbs=*
```

Hence the declarations in config/base CSV used to build the bundle isn't required at all, thus removing those.


vimalk78 (approved):
LGTM

PR #336, opened 2024-01-08
ci: improve model-server and estimator validation

sthaha:
Currently, CI fails because model-server image absent in CI and  validation does not wait enough for Model Server to be running. This is fixed by 
1. allowing additional  images to be loaded into cluster and using that to load model-server image
2. improving  the validation of model-server and estimator into 2 functions which provide each amble time to be up and running.


vimalk78 (approved):
/lgtm

PR #341, opened 2024-01-16
fix(controller): update status only if needed

sthaha:
This commit fixes kepler-internal controller to prevent infinite reconciliation caused by status updates even when there aren't any.


vprashar2929 (approved):
LGTM

PR #345, opened 2024-01-17
feat(dashboard): add detailed node power information

sthaha:
This commit adds a detailed node power information table to the overview dashboard. It also works around the OpenShift dashboard UI bug when rendering tables. The bug is that the first column must have unique values. This caused only a single CPU information to be rendered as `kepler_node_info` had components_power_source to be the first item in the list which had `estimator` as the only value.

A similar workaround is applied for "Detailed Node Information" table to have `instance` as the first field in the prometheus query result.

The issue is better explained in https://issues.redhat.com/browse/OU-309 

image


vprashar2929 (approved):
LGTM

PR #348, opened 2024-01-23
fix(test): accept deployment ns

sthaha:
This is required to support running tests when kepler is deployed to
non-default namespace. E.g. on OpenShift, it may be deployed onto a
different namespace than the default `kepler-operator`.


vprashar2929 (approved):
LGTM

PR #352, opened 2024-01-25
POWERMON-199: Use correct kepler namespace in must-gather

vimalk78:
Use labels to get correct namespace name.

### DIFF
--- a/must-gather/common
+++ b/must-gather/common
@@ -17,7 +17,7 @@
 
 set -eu -o pipefail
 
-export KEPLER_NS="kepler-operator"
+export KEPLER_NS=$(oc get ns -l app.kubernetes.io/managed-by=kepler-operator --output=jsonpath='{.items[*].metadata.name}')
 export LOGFILE_NAME="gather-debug.log"
 
 get_timestamp() {


must-gather/common
```diff
@@ -17,7 +17,7 @@
 
 set -eu -o pipefail
 
-export KEPLER_NS="kepler-operator"
+export KEPLER_NS=$(oc get ns -l app.kubernetes.io/managed-by=kepler-operator --output=jsonpath='{.items[*].metadata.name}')
```
sthaha, line 20: perhaps only the first item instead of all items in the list? 
`jsonpath={.items[0].metadata.name}` ?

sthaha (approved):
I think `-o name` may be

sthaha:
On a related topic, perhaps `status.deploymentNamespace` can be added to kepler.

must-gather/common
vimalk78: ideally there should be only 1.

sthaha: yes, indeed .. already approved.
In a different PR, I think it is better to iterate on all namespaces to capture the details. That way we get all namespaces (perhaps user has both community and power-mon installed) etc ...

sthaha (approved):

PR #357, opened 2024-02-01
feat: update uninstall script to capture operator name

vprashar2929:
This PR fetches the operator name`(power-monitoring-operator/kepler-operator)` from the cluster instead of the hard code name.


hack/uninstall-operator.sh
```diff
@@ -122,7 +122,11 @@ list_olm_resources() {
 }
 
 main() {
-	local operator="kepler-operator"
+	local operator=""
```
sthaha, line 125: lets rather make this a GLOBAL var and call a `init` function to initialize these globals which need to be calculated, after `parse_args`.

sthaha (approved):

PR #361, opened 2024-02-12
chore: update go to 1.21

sthaha:
Bumps up Go to 1.21 and updates modules. Importantly the `controller-runtime` https://github.com/kubernetes-sigs/controller-runtime/blob/main/go.mod#L3 which  uses 1.21


vprashar2929 (approved):
lgtm

Issue #202, opened 2023-09-04
Expose config for collecting cgroup and kubelet metrics in kepler exporter API


vimalk78:
enable/dsiable collecting cgroup and kubelet metrics by exposing bool  config params in Exporter API.

Ref: https://github.com/sustainable-computing-io/kepler/pull/876/files
        https://github.com/sustainable-computing-io/kepler/issues/901

vimalk78:
@rootfs @sunya-ch need your input here

sunya-ch:
Could you clarify more about this issue? Do you mean to expose it via the  Kepler command arguments not via the config map?

vimalk78:
i mean expose it in `ExporterSpec`

```
apiVersion: kepler.system.sustainable.computing.io/v1alpha1
kind: Kepler
metadata:
  labels:
    app.kubernetes.io/name: kepler
    app.kubernetes.io/instance: kepler
    app.kubernetes.io/part-of: kepler-operator
  name: kepler
spec:
  exporter:
    port: 9103
    enableCGroupMetrics: true
    enableKubeletMetrics: true
```
these values will be used to populate the kepler configmap

sunya-ch:
Thank you. The key of the configmap variables are all listed here: https://github.com/sustainable-computing-io/kepler/blob/02c16c1e223e32b1d1f20f3c7ad8e8959eb2fd87/pkg/config/config.go#L77.

sthaha:
@vimalk78 How about something like this. This allows us to add additional fields to each of the metric later on.  E.g. say in future, cgroup can not only be enabled but also a filter can be applied on the metrics, this would allow for change to be
added  without  breaking the api. Also grouping under metrics may allow us to add config that applies to all metrics (if any) in 
future without breaking api


```yaml
exporter:
  metrics:  # keep all metrics under MetricSpec
    cgroups:  # feature specific config
        enabled: true
    kubelet:
        enabled: true
    foobar:
       enabled: true
       additional: config
```

In general, I am wary of boolean flags.

sunya-ch:
In addition to `enabled`, we also need a flag for `exposed`. Sometimes, we enable it for power model estimation but does not want to expose it to prometheus.

vprashar2929:
closing this issue as it is not relevant according to latest code base

PR #364, opened 2024-03-02
POWERMON-235: Add support for Redfish

vimalk78:
This PR
 - Extends Kepler and KeplerInternal API to add `Exporter.Redfish`
 - Redfish spec has fields for adding reference to a secret which should
   contain Redfish credentials.
 - The user supplied secret is mounted in the kepler container, and daemon set spec is changed which causes kepler containers to restart.
 - Added Unit tests and e2e test


pkg/api/v1alpha1/kepler_internal_types.go
```diff
@@ -36,9 +36,15 @@ type InternalExporterDeploymentSpec struct {
...
 
+type InternalRedfishSpec struct {
+	RedfishSpec `json:",inline"`
+}
+
 type InternalExporterSpec struct {
 	// +kubebuilder:validation:Required
 	Deployment InternalExporterDeploymentSpec `json:"deployment"`
+
+	Redfish *InternalRedfishSpec `json:"redfish,omitempty"`
```
sunya-ch, line 47: We are going to extend spec for configuring the other metrics as well (such as to enable/disable cgroups metrics and so on) - check https://github.com/sustainable-computing-io/kepler-operator/issues/202. 
Thus, I think we should move Redfish to be under `metric.energy` and resource utilization metrics such as cGroups should go under `metric.resource`. 

Refer to terms in Kepler:
```golang
// https://github.com/sustainable-computing-io/kepler/blob/3d0bad753d68f74a0955f6e25d2a9d8e93507d08/pkg/collector/stats/stats.go#L47
type Stats struct {
	ResourceUsage map[string]*types.UInt64StatCollection
	EnergyUsage   map[string]*types.UInt64StatCollection
}
```

sunya-ch (commented):
I think this PR is worth merging for configuring the redfish. Please kindly find my comments above.

pkg/api/v1alpha1/kepler_internal_types.go
vimalk78: This seems to be a bigger change to the API. Can we do this api change in a separate PR?

sunya-ch: Sure. I think we can update that on a different PR. @sthaha What do you think?

sunya-ch:
This PR also made a change on Kepler resource. Shall we first limit this change to only KeplerInternal? 
Also, if `redfish.secretRef` is not set in the CR, it seems to make it not mounting since there is no default value for that.

vimalk78:
> Also, if `redfish.secretRef` is not set in the CR, it seems to make it not mounting

thats by design. once `secretRef` is set, then only `-redfish-cred-file-path` is added to container

pkg/components/exporter/exporter.go
```diff
@@ -45,6 +45,9 @@ const (
 	overviewDashboardName = "power-monitoring-overview"
 	nsInfoDashboardName   = "power-monitoring-by-ns"
 	DashboardNs           = "openshift-config-managed"
+	REDFISH_ARGS          = "-redfish-cred-file-path=/etc/redfish/redfish.csv"
+	REDFISH_CSV           = "redfish.csv"
+	REDFISH_ANNOTATION    = "kepler.system.sustainable.computing.io/redfish-secret-ref"
```
sthaha, line 50: could you please change this to follow the convention? 
If these constants aren't used outside the package, lets make them private.

pkg/controllers/kepler_internal.go
```diff
@@ -149,6 +179,24 @@ func (r KeplerInternalReconciler) getInternal(ctx context.Context, req ctrl.Requ
 	return &ki, nil
 }
 
+func (r KeplerInternalReconciler) getRedfishSecret(secretName string) (*corev1.Secret, error) {
+	redfishSecret := corev1.Secret{}
+	redFishSecretName := types.NamespacedName{
+		Namespace: KeplerDeploymentNS,
```
sthaha, line 185: shouldn't this be the namespace that you get from the `kepler-internal` instead of the deployment namespace of `kepler` ? 
As a rule of thumb, internal must not use (or have any knowledge) of `Kepler` related config or variables.

pkg/controllers/kepler_internal.go
```diff
@@ -490,16 +538,79 @@ func exporterReconcilers(ki *v1alpha1.KeplerInternal, cluster k8s.Cluster) []rec
...
+type KeplerDaemonSetReconciler struct {
+	ki      v1alpha1.KeplerInternal
+	updater reconcileFn
+	ds      *appsv1.DaemonSet
+}
+
+func (r KeplerDaemonSetReconciler) Reconcile(ctx context.Context, cli client.Client, s *runtime.Scheme) reconciler.Result {
+
+	secretRef := r.ki.Spec.Exporter.Redfish.SecretRef
+	secret, err := r.getRedfishSecret(ctx, cli, secretRef)
```
sthaha, line 577: I think we should fail reconcile, set the status to available false,  if a secret is specified and is missing or invalid (missing the redfish.csv) . Otherwise there is no way for user to know if kepler is using redfish or not.

sthaha (changes requested):

sthaha:
@vimalk78 nice work overall. It worked pretty well in my test. But few comments to address. 
The main one being that secret must be present for kepler deployment to continue and the reason for failure to deploy must be reported back in `Availability` status.

pkg/controllers/kepler_internal.go
```diff
@@ -74,12 +77,39 @@ func (r *KeplerInternalReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rbacv1.ClusterRoleBinding{}, genChanged).
 		Owns(&rbacv1.ClusterRole{}, genChanged)
 
+	c = c.Watches(&corev1.Secret{},
+		handler.EnqueueRequestsFromMapFunc(r.mapSecretToRequests))
```
sthaha, line 81: Also, handler.ResourceVersionChanged{}

pkg/controllers/kepler_internal.go
```diff
@@ -74,12 +77,39 @@ func (r *KeplerInternalReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// mapSecretToRequests returns the reconcile requests for kepler-internal objects for which an associated redfish secret has been changed.
+func (r *KeplerInternalReconciler) mapSecretToRequests(ctx context.Context, object client.Object) []reconcile.Request {
+
+	secret, ok := object.(*corev1.Secret)
+	if !ok {
+		return nil
+	}
+
+	ks := v1alpha1.KeplerInternalList{}
+	if err := r.List(ctx, &ks); err != nil {
```
sthaha, line 98: I think we can improve the perf a bit more by using the `fieldSelector`
see: https://book.kubebuilder.io/reference/watching-resources/externally-managed

pkg/controllers/kepler_internal.go
```diff
@@ -74,12 +77,39 @@ func (r *KeplerInternalReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	return c.Complete(r)
 }
 
+// mapSecretToRequests returns the reconcile requests for kepler-internal objects for which an associated redfish secret has been changed.
+func (r *KeplerInternalReconciler) mapSecretToRequests(ctx context.Context, object client.Object) []reconcile.Request {
+
+	secret, ok := object.(*corev1.Secret)
+	if !ok {
+		return nil
+	}
```
sthaha, line 95: This should not be needed since the watch is already set for `v1.Secret`

config/rbac/role.yaml
```diff
@@ -9,6 +9,7 @@ rules:
   resources:
   - daemonsets
   - deployments
+  - secrets
```
sthaha, line 12: we don't create secrets do we ?

pkg/components/exporter/exporter.go
```diff
@@ -132,6 +137,19 @@ func NewDaemonSet(detail components.Detail, k *v1alpha1.KeplerInternal) *appsv1.
...
+	spec := ds.Spec.Template.Spec
+	spec.Containers[IdxKeplerContainer].Command = append(spec.Containers[IdxKeplerContainer].Command, REDFISH_ARGS)
+	spec.Containers[IdxKeplerContainer].VolumeMounts = append(spec.Containers[IdxKeplerContainer].VolumeMounts,
+		corev1.VolumeMount{Name: "redfish-cred", MountPath: "/etc/redfish", ReadOnly: true})
+	spec.Volumes = append(spec.Volumes,
+		k8s.VolumeFromSecret("redfish-cred", secret.ObjectMeta.Name))
+	ds.Spec.Template.Spec = spec
+	ds.Spec.Template.Annotations = map[string]string{
+		REDFISH_ANNOTATION: secret.ResourceVersion,
+	}
```
sthaha, line 150: I think it may be sufficient to just annotate the ds instead of the template. Lets handle this in a different PR.

pkg/controllers/kepler_internal.go
```diff
@@ -46,12 +48,12 @@ type KeplerInternalReconciler struct {
...
 
 // RBAC for running Kepler exporter
-//+kubebuilder:rbac:groups=apps,resources=daemonsets;deployments,verbs=list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps,resources=daemonsets;deployments;secrets,verbs=list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=list;watch;create;update;patch;delete;use
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;prometheusrules,verbs=list;watch;create;update;patch;delete
 
 // RBAC required by Kepler exporter
-//+kubebuilder:rbac:groups=core,resources=nodes/metrics;nodes/proxy;nodes/stats,verbs=get;list;watch
+//+kubebuilder:rbac:groups=core,resources=nodes/metrics;nodes/proxy;nodes/stats;secrets,verbs=get;list;watch
```
sthaha, line 56: This shouldn't be necessary since Kepler exporter does not watch or create secrets.

tests/run-e2e.sh
```diff
@@ -302,7 +302,7 @@ restart_operator() {
 	info "scale down Operator"
 	run kubectl scale -n "$OPERATORS_NS" --replicas=0 "$deployment"
 	run kubectl wait -n "$OPERATORS_NS" --for=delete \
-		pods -l app.kubernetes.io/component=operator --timeout=60s
+		pods -l app.kubernetes.io/component=manager --timeout=60s
```
sthaha, line 305: good find!

sthaha (approved):

PR #365, opened 2024-03-03
WIP basic structure for adding Related Objects.

KaiyiLiu1234:
Structure was taken from ClusterOperatorStatus. Populating Related Objects should occur in Reconcile alongside updateStatus. Reference https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#specDescriptors-1 for xdescriptors that can be used to portray status (podstatuses or with text).


sthaha:
Closing this as this isn't a supported feature for all operators.

PR #371, opened 2024-03-11
fix(redfish): skip internals without redfish secret

sthaha:


bundle/manifests/kepler-operator.clusterserviceversion.yaml
```diff
@@ -378,7 +378,7 @@ spec:
   relatedImages:
   - image: quay.io/sustainable_computing_io/kepler:release-0.7.2
     name: kepler
-  replaces: kepler-operator.v0.9.2
+  replaces: kepler-operator.v0.0.0-e2e
```
vimalk78, line 381: need to revert

vimalk78 (approved):

PR #385, opened 2024-04-02
ci: setup kind without monitoring

sthaha:
Removing monitoring reduces the disk usage. 
Operator requires servicemonitor and prometheusrule crd for installing kepler thus these crds need to be added to the cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ docs: crdoc manifests ## Generate docs.
 CLUSTER_PROVIDER ?= kind
 LOCAL_DEV_CLUSTER_VERSION ?= main
 GRAFANA_ENABLE ?= true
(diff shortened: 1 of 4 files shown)


vimalk78 (approved):
/lgtm

Makefile
```diff
@@ -118,13 +118,15 @@ docs: crdoc manifests ## Generate docs.
 CLUSTER_PROVIDER ?= kind
 LOCAL_DEV_CLUSTER_VERSION ?= main
 GRAFANA_ENABLE ?= true
```
vprashar2929, line 120: Grafana enable will deploy prometheus by default

sthaha: +1 .. For this PR, it disables Monitoring in CI. 
I will fix the makefile issue  in a new pr #386

PR #393, opened 2024-05-13
fix: allow only overview dashboard to be visible inside OpenShift developer console

vprashar2929:
This PR restricts visibility to only the power-monitoring-overview dashboard in the OpenShift developer console due to the
console's limitation on per-namespace visualization

### DIFF
--- a/pkg/components/exporter/exporter.go
+++ b/pkg/components/exporter/exporter.go
@@ -164,8 +164,7 @@ func openshiftDashboardObjectMeta(name string) metav1.ObjectMeta {
 		Name:      name,
 		Namespace: DashboardNs,
 		Labels: components.CommonLabels.Merge(k8s.StringMap{
-			"console.openshift.io/dashboard":     "true",
-			"console.openshift.io/odc-dashboard": "true",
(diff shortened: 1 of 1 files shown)


pkg/components/exporter/exporter.go
```diff
@@ -160,13 +160,18 @@ func MountRedfishSecretToDaemonSet(ds *appsv1.DaemonSet, secret *corev1.Secret,
...
 
 func openshiftDashboardObjectMeta(name string) metav1.ObjectMeta {
+	var dashboardLabels k8s.StringMap = k8s.StringMap{
+		"console.openshift.io/dashboard": "true",
+	}
+
+	// NOTE ommiting the below label when configuring the power-monitoring-by-ns dashboard
+	if name != nsInfoDashboardName {
+		dashboardLabels["console.openshift.io/odc-dashboard"] = "true"
+	}
```
sthaha, line 170: We should rather move this code to Overview Dashboard.

sthaha (approved):

sthaha (approved):

PR #395, opened 2024-05-21
fix(OLM): ensure kepler can be created through Console

sthaha:
Previously, when creating Kepler instance through OpenShift Console / UI, the redfish spec which is optional also had its fields initialized as well.  Notably `redfish.secrefRef` was initialized to "". This prevented creation of `kepler` instance since the `secretRef` is a required field.

This commit fixes the issue by marking `redfish` as `optional` and updating the default sample to reflect that.


vimalk78 (approved):
/lgtm

Issue #399, opened 2024-06-03
Disable EXPOSE_CGROUP_METRICS in Kepler config


vprashar2929:
Since the collection of CGROUP metrics is not required anymore by the latest Kepler and it also increases the overhead in terms of CPU and memory utilization for Kepler pods it's best to set the EXPOSE_CGROUP_METRICS to false

vprashar2929:
closing as #400  is merged

PR #400, opened 2024-06-04
fix: disable cgroup metrics for Kepler

vprashar2929:
Sets `EXPOSE_CGROUP_METRICS=false` in Kepler's config. This change addresses issue #399


pkg/components/exporter/exporter.go
```diff
@@ -257,7 +257,7 @@ func NewConfigMap(d components.Detail, k *v1alpha1.KeplerInternal) *corev1.Confi
 		"EXPOSE_HW_COUNTER_METRICS":  "true",
 		"EXPOSE_IRQ_COUNTER_METRICS": "true",
 		"EXPOSE_KUBELET_METRICS":     "true",
-		"EXPOSE_CGROUP_METRICS":      "true",
+		"EXPOSE_CGROUP_METRICS":      "false",
```
sthaha, line 260: https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/pkg/components/exporter/exporter.go#L620 Ahem ahem :p

sthaha (changes requested):
more than one place to fix :)

pkg/components/exporter/exporter.go
sthaha: Also, I think we can completely remove this from the configmap
cc: @marceloamaral ?

vprashar2929: I think by default Kepler takes value of `ENABLE_EBPF_CGROUPID` as [true](https://github.com/sustainable-computing-io/kepler/blob/29d0b00f2362ac28d14b887af9a3eb6140f281b0/pkg/config/config.go#L70) and `EXPOSE_CGROUP_METRICS` as [true](https://github.com/sustainable-computing-io/kepler/blob/29d0b00f2362ac28d14b887af9a3eb6140f281b0/pkg/config/config.go#L77)

pkg/components/exporter/exporter.go
```diff
@@ -617,7 +617,7 @@ func newExporterContainer(kiName, dsName string, deployment v1alpha1.InternalExp
 		Command: []string{
 			"/usr/bin/kepler",
 			"-address", bindAddress,
-			"-enable-cgroup-id=true",
+			"-enable-cgroup-id=$(ENABLE_EBPF_CGROUPID)",
 			"-enable-gpu=$(ENABLE_GPU)",
```
sthaha, line 621: can we not remove both since the config already has this value?

sthaha (changes requested):

sthaha (approved):

Issue #381, opened 2024-03-19
.spec.volumes: field not declared in schema


jtaleric:
Attempting to deploy, with 

Following the docs to deploy the operator, I ran : 

```
$  make deploy OPERATOR_IMG=quay.io/sustainable_computing_io/kepler-operator:v1alpha1
```

```
./hack/tools.sh kustomize
   ✅ kustomize matching v3.8.7 already installed
{Version:kustomize/v3.8.7 GitCommit:ad092cc7a91c07fdf63a2e4b7f13fa588a39af4f BuildDate:2020-11-11T23:14:14Z GoOs:linux GoArch:amd64}
./hack/tools.sh controller-gen
   ✅ controller-gen matching Version: v0.12.1 already installed
Version: v0.12.1
/home/jtaleric/code/kepler-operator/tmp/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./pkg/..."
/home/jtaleric/code/kepler-operator/tmp/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./pkg/..." output:crd:artifacts:config=config/crd/bases
/home/jtaleric/code/kepler-operator/tmp/bin/kustomize build config/crd | \
	kubectl apply --server-side --force-conflicts -f -
customresourcedefinition.apiextensions.k8s.io/keplerinternals.kepler.system.sustainable.computing.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/keplers.kepler.system.sustainable.computing.io serverside-applied
/home/jtaleric/code/kepler-operator/tmp/bin/kustomize build config/default | \
	sed  -e "s|<OPERATOR_IMG>|quay.io/sustainable_computing_io/kepler-operator:v1alpha1|g" \
	     -e "s|<KEPLER_IMG>|quay.io/sustainable_computing_io/kepler:release-0.7.2|g" \
	| tee tmp/deploy.yaml | \
	kubectl apply --server-side --force-conflicts -f -
namespace/kepler-operator-system serverside-applied
customresourcedefinition.apiextensions.k8s.io/keplerinternals.kepler.system.sustainable.computing.io serverside-applied
customresourcedefinition.apiextensions.k8s.io/keplers.kepler.system.sustainable.computing.io serverside-applied
serviceaccount/kepler-operator-controller-manager serverside-applied
role.rbac.authorization.k8s.io/kepler-operator-leader-election-role serverside-applied
[...]

PR #401, opened 2024-06-06
fix: make deploy target

vprashar2929:
This PR address the issue #381 by:

* Enabling cert-manager that is necessary by webhooks during the
      deployment of Operator on k8s.

* Added a separate config for k8s that contains the manifests that
   are specific for deploying on k8s.
   * `config/default` points to OpenShift specific manifests
   * `config/k8s` points to manifests specific for k8s
    
* Added overlays in manager to isolate k8s and
      OpenShift


bundle/manifests/kepler.system.sustainable.computing.io_keplers.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    cert-manager.io/inject-ca-from: kepler-operator-system/kepler-operator-serving-cert
```
sthaha, line 5: How does this behave on OpenShift that has cert-manager installed?

Makefile
```diff
@@ -245,7 +248,10 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 		kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
-deploy: install ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+deploy: install cmctl ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	kubectl apply --server-side --force-conflicts -f \
+		https://github.com/jetstack/cert-manager/releases/download/v$(CERTMANAGER_VERSION)/cert-manager.yaml
```
sthaha, line 253: Imagine running this on k8s with a cert-manager already installed and configured? Can this potentially break their system? 

IMO, we should first validate if certmanager exists and install it only if there isn't any. 
Additionally, consider exiting with error if cert-manager is absent. 
We can also provide `install-certmanager` helper to make life easier for vanilla k8s installation.

hack/cluster.sh
```diff
@@ -31,7 +31,6 @@ declare -r PROJECT_ROOT
 declare -r TMP_DIR="$PROJECT_ROOT/tmp"
 declare -r DEV_CLUSTER_DIR="$TMP_DIR/local-dev-cluster"
 declare -r BIN_DIR="$TMP_DIR/bin"
-declare -r OPERATOR_SDK_VERSION=${OPERATOR_SDK_VERSION:-v1.27.0}
```
sthaha, line 34: any reason to remove this?

sthaha (commented):
Thanks a lot @vprashar2929 !

hack/cluster.sh
vprashar2929: wasn't used in the script anywhere

config/default/manager_config_patch.yaml
vprashar2929, line 1: Deleted this file as it wasn't used anywhere

PR #403, opened 2024-06-11
fix(cgroup): enable cgroup without exposing cgroup_metrics

sthaha:
Cgroup is still used. Only the metrics are not going to be exposed anymore.

### DIFF
--- a/pkg/components/exporter/exporter.go
+++ b/pkg/components/exporter/exporter.go
@@ -253,7 +253,7 @@ func NewConfigMap(d components.Detail, k *v1alpha1.KeplerInternal) *corev1.Confi
 		"BIND_ADDRESS":               bindAddress,
 		"ENABLE_GPU":                 "true",
 		"ENABLE_QAT":                 "false",
-		"ENABLE_EBPF_CGROUPID":       "false",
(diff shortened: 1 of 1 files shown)


pkg/components/exporter/exporter.go
```diff
@@ -253,11 +253,11 @@ func NewConfigMap(d components.Detail, k *v1alpha1.KeplerInternal) *corev1.Confi
 		"BIND_ADDRESS":               bindAddress,
 		"ENABLE_GPU":                 "true",
 		"ENABLE_QAT":                 "false",
-		"ENABLE_EBPF_CGROUPID":       "false",
+		"ENABLE_EBPF_CGROUPID":       "true",
 		"EXPOSE_HW_COUNTER_METRICS":  "true",
 		"EXPOSE_IRQ_COUNTER_METRICS": "true",
 		"EXPOSE_KUBELET_METRICS":     "true",
-		"EXPOSE_CGROUP_METRICS":      "false",
+		`EXPOSE_CGROUP_METRICS`:      "false",
```
vprashar2929, line 260: Shouldn't this be in `""`?

vprashar2929 (approved):

PR #404, opened 2024-06-14
feat: migrate operator code base to support new operator-sdk version

vprashar2929:
This PR updates the code base for compatibility with the latest operator-sdk version(v1.35.0)

Key Changes:
* Moved API files from `pkg/api` to `api` folder
* Relocated `main.go` from `cmd/manager/main.go` to `cmd/main.go`
* Shifted controllers from `pkg/controllers` to `controllers` folder
* Made minor updates to components, utils and tests to use correct import paths
* Updated scaffolding config to align with the latest operator-sdk version
* Revised Operator, bundle and e2e Dockerfiles
* Upgraded operator-sdk to v1.35.0 and controller-gen to v0.13.0
* Updated bundle manifests accordingly


bundle/manifests/kepler-operator.clusterserviceversion.yaml
```diff
@@ -28,14 +28,14 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring
     containerImage: quay.io/sustainable_computing_io/kepler-operator:0.13.0
-    createdAt: "2024-05-22T07:06:13Z"
+    createdAt: "2024-06-24T08:00:23Z"
     description: 'Deploys and Manages Kepler on Kubernetes '
-    operators.operatorframework.io/builder: operator-sdk-v1.27.0
+    operators.operatorframework.io/builder: operator-sdk-v1.34.2
```
sthaha, line 33: How about we move to 1.35 ?

internal/controller/kepler_controller.go
vprashar2929, line 1: AFAIU we can opt-out the internal folder as all the controller code should be in the project root dir https://sdk.operatorframework.io/docs/building-operators/golang/migration/#controllers 
Any thoughts?

bundle/tests/scorecard/config.yaml
```diff
@@ -8,7 +8,7 @@ stages:
   - entrypoint:
     - scorecard-test
     - basic-check-spec
-    image: quay.io/operator-framework/scorecard-test:v1.25.1
+    image: quay.io/operator-framework/scorecard-test:v1.34.2
```
sthaha, line 11: 1.35.0 ? :)

hack/bundle.sh
```diff
@@ -58,7 +58,7 @@ main() {
 	info "Found old version: $old_bundle_version"
 
 	info "Building bundle version $VERSION"
-	run operator-sdk generate kustomize manifests --apis-dir=./pkg/api --verbose
+	run operator-sdk generate kustomize manifests --apis-dir=./api --verbose
```
sthaha, line 61: IIRC, api is the default dir. Could you please check?

internal/controller/kepler_controller.go
sthaha: Lets not keep the controller internal. `internal` packages have special meaning - https://go.dev/doc/go1.4#internalpackages

Dockerfile
```diff
@@ -12,28 +12,32 @@ COPY go.sum go.sum
...
-COPY must-gather/ must-gather/
-COPY cmd/ cmd/
+COPY cmd/main.go cmd/main.go
+COPY api/ api/
 COPY pkg/ pkg/
+COPY controllers/ controllers/
+
 
+# Copy must-gather scripts
+COPY must-gather/ must-gather/
```
sthaha, line 22: lets preserve the order based on chance of modifying these files.

config/webhook/service.yaml
```diff
@@ -19,3 +19,6 @@ spec:
     app.kubernetes.io/component: manager
     app.kubernetes.io/instance: controller-manager
     app.kubernetes.io/part-of: kepler-operator
+
+  # selector:
+  #   control-plane: controller-manager
```
sthaha, line 24: delete

PROJECT
```diff
@@ -9,6 +13,7 @@ repo: github.com/sustainable.computing.io/kepler-operator
 resources:
 - api:
     crdVersion: v1
+    namespaced: true
```
sthaha, line 16: this shouldn't marked as namespaced. Our APIs are cluster-scoped.

config/rbac/service_account.yaml
```diff
@@ -3,7 +3,7 @@ kind: ServiceAccount
 metadata:
   labels:
     app.kubernetes.io/name: serviceaccount
-    app.kubernetes.io/instance: controller-manager
+    app.kubernetes.io/instance: controller-manager-sa
```
sthaha, line 6: Have you tested upgrade path with this? Will there be 2 SA after we upgrade ?

vprashar2929: When the Operator is upgraded it doesn't change the name of the SA as the name is fixed to `controller-manager`

sthaha (approved):

PR #410, opened 2024-07-16
POWERMON-306: fix wrong Prometheus rule reference for OTHER Power Consumption

vprashar2929:
This commit fixes the issue where the wrong Prometheus rule was referenced for displaying `OTHER Power Consumption(W) by Pods` in the dashboard

### DIFF
--- a/hack/dashboard/assets/kepler/dashboard.json
+++ b/hack/dashboard/assets/kepler/dashboard.json
@@ -1100,7 +1100,7 @@
         {
           "datasource": "prometheus",
           "editorMode": "code",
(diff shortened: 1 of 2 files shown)


vprashar2929:
Verified the same on OpenShift

Screenshot 2024-07-16 at 6 14 04 PM

sthaha (changes requested):
Thanks for the PR @vprashar2929 .  Requesting changes to

*  refer to the original issue being fixed in the commit msg
* remove backticks from commit message

sthaha (approved):

Issue #420, opened 2024-08-21
Bump up model-server version


vprashar2929:
With the latest model-server release https://github.com/sustainable-computing-io/kepler-model-server/releases/tag/v0.7.11-2-release we should bump up the model-server version so that the operator also deploys the same

vprashar2929:
closing as #421 is merged

PR #421, opened 2024-08-22
chore: Bump up model-server to v0.7.11-2

vprashar2929:
This commit bumps up model-server to `v0.7.11-2`

Addresses: #420

### DIFF
--- a/pkg/components/estimator/estimator.go
+++ b/pkg/components/estimator/estimator.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	// NOTE: update tests/images.yaml when changing this image
-	StableImage          = "quay.io/sustainable_computing_io/kepler_model_server:v0.7.11"
+	StableImage          = "quay.io/sustainable_computing_io/kepler_model_server:v0.7.11-2"
 	waitForSocketCommand = "until [ -e /tmp/estimator.sock ]; do sleep 1; done && %s"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/components/estimator/estimator.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	// NOTE: update tests/images.yaml when changing this image
-	StableImage          = "quay.io/sustainable_computing_io/kepler_model_server:v0.7.11"
+	StableImage          = "quay.io/sustainable_computing_io/kepler_model_server:v0.7.11-2"
```
sthaha, line 31: @vprashar2929 we should also change the entrypoint and command
Please check the model-server compose files.

vprashar2929: Yeah forgot about that 😅

pkg/components/estimator/estimator.go
```diff
@@ -76,8 +76,8 @@ func Container(image string) corev1.Container {
 		ImagePullPolicy: corev1.PullIfNotPresent,
 		Name:            "estimator",
 		VolumeMounts:    mounts,
-		Command:         []string{"python3.10"},
-		Args:            []string{"-u", "src/estimate/estimator.py"},
+		Command:         []string{"estimator"},
+		Args:            []string{"-l", "info"},
```
sthaha, line 80: `-l info` is optional btw ...

sthaha (changes requested):
We should also update kepler: command to wait for both model-server and estimator

sthaha (approved):

Issue #429, opened 2024-10-09
Remove the steps for cleaning the runner when running e2e


vprashar2929:
Since we have moved to self hosted runners we should remove the steps from the CI and e2e script for cleaning the space as it is unnecessary now

vprashar2929:
closing as #430 got merged

PR #430, opened 2024-10-10
chore(ci): remove space reclamation steps for GH runners

vprashar2929:
This commit removes the steps previously used to reclaim space on GH runners. Since we have moved to self-hosted runners with more resources available, these steps are no longer necessary

Addresses: #429

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -134,7 +134,7 @@ jobs:
       KUBECONFIG: /tmp/.kube/config
       KIND_WORKER_NODES: 2
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-8-cores
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -165,7 +165,7 @@ jobs:
       - name: Run Operator Upgrade
         shell: bash
         run: |
-          ./tests/run-e2e.sh --ci
+          ./tests/run-e2e.sh
```
sthaha, line 168: Lets preserve `--ci` since the flag can be used to decide on other things than run `container rm` later.

.github/workflows/pr-checks.yaml
```diff
@@ -224,18 +224,9 @@ jobs:
...
-          sudo rm -rf /var/lib/apt/lists/*
-          sudo rm -rf /tmp/*
-          sudo rm -rf /opt/hostedtoolcache/CodeQL
-          sudo rm -rf /opt/hostedtoolcache/Python
-          df -kh
-
       - name: Run e2e tests
         run: |
-          ./tests/run-e2e.sh --ci --no-upgrade
+          ./tests/run-e2e.sh --no-upgrade
```
sthaha, line 229: same here .. lets keep `--ci`  🙏

sthaha (changes requested):
Requesting changes to keep the `--ci`

sthaha (approved):

PR #469, opened 2025-04-22
feat: expose version info

vprashar2929:
This commit adds the support to log the version related information in manager's logs. This is useful for debugging purposes. The version information consists of the following:
- Operator version (from the VERSION file)
- Build time
- Git Branch
- Git Commit
- GOOS
- GOARCH


vprashar2929:
The downside of running `make build` instead of `go build ....` in the Dockerfile is the increase in build time as `make build` depends upon other make targets, which also download/install some dependent tools.
In my comparisons, the original build time is ~69s, and with this chang,e it will be ~130s 

No change in image size

pkg/version/version.go
```diff
@@ -0,0 +1,35 @@
+package version
```
sthaha, line 1: missing license

sthaha (approved):

pkg/version/version.go
vprashar2929: my bad 😞 I will add license checks next

PR #477, opened 2025-04-25
chore: update must-gather to support power-monitor

vprashar2929:
This commit updates the must-gather script to also gather power-monitor related information. It refactors the existing gather script to include support for both Kepler and PowerMonitor making sure that the gather script is compatible with both instance types

### DIFF
--- a/must-gather/gather
+++ b/must-gather/gather
@@ -13,7 +13,8 @@ declare OPERATOR=""
 declare COLLECTION_DIR=""
 
 declare OLM_INFO_DIR=""
-declare KEPLER_NS=""
+declare INSTANCE_NS=""
+declare INSTANCE_TYPE=""
 declare KEPLER_OPERATOR_INFO_DIR=""
 
 declare UWM_INFO_DIR=""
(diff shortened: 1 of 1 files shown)


sthaha (commented):
@KaiyiLiu1234 could you please help review this?

KaiyiLiu1234:
Implementation looks good. I still need to perform manual validation on openshift local.

KaiyiLiu1234:
@vprashar2929 Tested on Openshift cluster and everything works well. must-gather runs without error. must-gather directory contains power-monitor related events, kubernetes objects yaml, relevant logs from power-monitor deployed pod(s). My only concern is that power-monitor-info./ebpf-info, power-monitor-info./kernel-ebpf-features, and power-monitor-info./node-cpuid-info are empty. I think ideally for power-monitor-info these should just be removed as i assume these logs are more for original kepler. This will be up to your discretion.

KaiyiLiu1234 (approved):
No issues apart from the empty logs described in the previous comment.

vprashar2929:
Yeah I agree. My thinking is that once we remove Kepler(old) will remove everything we capture related to that.

KaiyiLiu1234:
That sounds good to me!

sthaha:
> Yeah I agree. My thinking is that once we remove Kepler(old) will remove everything we capture related to that.

@vprashar2929 could you please add a tracker issue for this? 

Merging the PR based on @KaiyiLiu1234 's comment.

PR #480, opened 2025-04-28
feat(dashboards): Add Power Monitor Info/Node OCP Dashboards

KaiyiLiu1234:
Operator now deploys Openshift Dashboards for Power Monitor CR when on an OCP cluster. Added Power Monitor Node OCP Dashboard which shows 4 energy watt dashboards (pkg, core, dram, uncore). Added Power Monitor Info OCP Dashboard which shows kepler build info and cpu info.


KaiyiLiu1234:
@vprashar2929 @sthaha what are your thoughts on instead having a single dashboard and a variable which lets you switch between the available rapl zones? Currently, it is 4 hard coded dashboards (1 for pkg, core, uncore, dram).

pkg/components/power-monitor/deployment.go
```diff
@@ -126,6 +133,14 @@ func NewPowerMonitorService(pmi *v1alpha1.PowerMonitorInternal) *corev1.Service
 	}
 }
 
+func NewPowerMonitorNodeDashboard(d components.Detail) *corev1.ConfigMap {
```
sthaha, line 136: we need tests :)

internal/controller/power_monitor_internal.go
```diff
@@ -153,6 +153,23 @@ func openshiftPowerMonitorClusterResources(pmi *v1alpha1.PowerMonitorInternal, c
...
 }
 
+func openshiftPowerMonitorNamespacedResources(pmi *v1alpha1.PowerMonitorInternal, cluster k8s.Cluster) []client.Object {
+	oshift := pmi.Spec.OpenShift
+
+	if cluster != k8s.OpenShift || !oshift.Enabled {
+		return nil
+	}
+
+	res := []client.Object{}
```
sthaha, line 163: question: is it possible to return real objects instead of client.Object?

sthaha (changes requested):
LGTM! Requesting changes to add tests.

sthaha:
@vprashar2929 , could you please help validate this manually?

sthaha:
> @vprashar2929 @sthaha what are your thoughts on instead having a single dashboard and a variable which lets you switch between the available rapl zones? Currently, it is 4 hard coded dashboards (1 for pkg, core, uncore, dram).

We should be able to use `sum by (name) (kepler_node_rapl_zone)` now (available in latest reboot) label `name` to get a list of all available zones.

KaiyiLiu1234:
> We should be able to use `sum by (name) (kepler_node_rapl_zone)` now (available in latest reboot) label `name` to get a list of all available zones.

sounds good. next pr will include this update

internal/controller/power_monitor_internal.go
```diff
@@ -181,15 +197,15 @@ func powerMonitorExporters(pmi *v1alpha1.PowerMonitorInternal, cluster k8s.Clust
 		powermonitor.NewPowerMonitorServiceAccount(pmi),
 		powermonitor.NewPowerMonitorService(pmi),
 		powermonitor.NewPowerMonitorServiceMonitor(pmi),
-		// powermonitor.NewPowerMonitorPrometheusRule(kx), do not include until metrics are made available
+		// powermonitor.NewPowerMonitorPrometheusRule(kx), prometheus rule is not necessary at the moment
```
vprashar2929, line 200: We have to update all our dashboard queries anyway when we enable rules, even if they are node level ones.🤔 Have to validate against a long-running cluster(Kepler running more than 2 days) to see how the queries respond

pkg/components/power-monitor/assets/dashboards/power-monitoring-info.json
```diff
@@ -0,0 +1,378 @@
...
+      ],
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "To6-2So4k"
+          },
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "group(kepler_build_info{container=~\"power-monitor\"}) by (arch, instance, goversion, prometheus, revision, version)",
```
vprashar2929, line 169: how about `max(kepler_build_info{container=~\"power-monitor\"}) by (arch, goversion, revision, version)`? Since the same version of Kepler will be deployed on all the nodes, I don't think showing per-instance info makes sense here. A single row should be sufficient. WDYT?

pkg/components/power-monitor/assets/dashboards/power-monitoring-info.json
```diff
@@ -0,0 +1,378 @@
...
+        }
+      ],
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "group(kepler_cpu_info{container=~\"power-monitor\"}) by (core_id, instance, model_name, physical_id, pod, processor, vendor_id)",
```
vprashar2929, line 328: How about we only show Instance, Model name, VendorID and Cores? Maybe in another PR

pkg/components/power-monitor/assets/dashboards/power-monitoring-by-node.json
```diff
@@ -0,0 +1,478 @@
...
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "kepler_node_package_watts{container=~\"power-monitor\", instance=~\"$instance\"}",
+          "hide": false,
+          "interval": "",
+          "legendFormat": "(PKG) - {{instance}}",
```
vprashar2929, line 90: how about `(Package) - {{instance}}`?

sthaha:
@KaiyiLiu1234, could you also please 
* add screenshots to the PR. 
* squash commits or spilt commits to logical units.

internal/controller/power_monitor_internal.go
KaiyiLiu1234: The reason it returns client.Object is because openshiftnamespacedresources can in theory return a variety of different objects. If that is too generic, we can instead divide it into multiple different functions. That said, it does not change much because resourceReconcilers accepts a list of client.Object as well. A lot of the reusable functions used in the operator are generic and so accept type client.Object.

internal/controller/power_monitor_internal.go
KaiyiLiu1234: understood. currently, dashboard queries are just the queries produced by kepler. I do not know if we will even need prometheus rules. The previous rules produced by old kepler were very confusing because they involved time ranges (ex. last 24 hours). I think if users want a certain type of metric, they should let us know first.

pkg/components/power-monitor/assets/dashboards/power-monitoring-info.json
KaiyiLiu1234: Sure! I can make this change now.

pkg/components/power-monitor/assets/dashboards/power-monitoring-info.json
KaiyiLiu1234: I think this makes sense for now. I imagine when we have multinode clusters with different arch, we can get a scenario where kepler_build_info has multiple values with different arch. But at the moment I do not think we will in the medium future have to handle that so I agree with using the metric you provided.

KaiyiLiu1234:
> > @vprashar2929 @sthaha what are your thoughts on instead having a single dashboard and a variable which lets you switch between the available rapl zones? Currently, it is 4 hard coded dashboards (1 for pkg, core, uncore, dram).
> 
> We should be able to use `sum by (name) (kepler_node_rapl_zone)` now (available in latest reboot) label `name` to get a list of all available zones.

then in that case, should we make that a new PR where the zone will be added as a variable and we can select which zone to view?

sthaha:
@KaiyiLiu1234, could you also please squash the commits if the PR is good to final review?

KaiyiLiu1234:
PR Dashboard Results:
Screenshot From 2025-05-04 15-20-14
Screenshot From 2025-05-04 15-20-28
Screenshot From 2025-05-04 15-20-38

KaiyiLiu1234:
@sthaha Squashed commits. The current commits are atomic and should not be merged together.

KaiyiLiu1234:
@vprashar2929 Ready for testing whenever you can!

internal/controller/power_monitor_internal.go
vprashar2929: yeah I am sure this will come as a requirement in future. But for now, we can go ahead with plain promql

pkg/components/power-monitor/assets/dashboards/power-monitoring-info.json
```diff
@@ -0,0 +1,407 @@
...
+          "exemplar": false,
+          "expr": "max(kepler_build_info{container=~\"power-monitor\"}) by (arch, goversion, revision, version)",
+          "format": "table",
+          "instant": true,
+          "legendFormat": "",
+          "range": false,
+          "refId": "A"
+        }
+      ],
+      "title": "Node - Kepler Build Info",
```
vprashar2929, line 144: `"Kepler Build Info"` ?

vprashar2929 (dismissed):

KaiyiLiu1234:
@sthaha Rebased to 3 commits. Fourth commit was to resolve a merge conflict after force pushing (cannot rebase that).

sthaha:
@vprashar2929 It seems we are skipping required tests. Do you know why? 

image

sthaha (approved):

vprashar2929:
my bad 😞  I missed a check in workflow. Added now: #496 Thanks for pointing that 🙇

PR #481, opened 2025-04-29
tests(power-monitor): Add initial setup for power-monitor e2e tests

KaiyiLiu1234:
Added e2e test framework for testing power-monitor and power-monitor-internal reconciliation.

### DIFF
--- a/tests/e2e/main_test.go
+++ b/tests/e2e/main_test.go
@@ -13,12 +13,14 @@ import (
 )
 
 const (
-	keplerImage = `quay.io/sustainable_computing_io/kepler:release-0.7.12`
+	keplerImage       = `quay.io/sustainable_computing_io/kepler:release-0.7.12`
+	keplerRebootImage = `quay.io/sustainable_computing_io/kepler-reboot:v0.0.4`
 )
 
 var (
-	Cluster         k8s.Cluster = k8s.Kubernetes
-	testKeplerImage string
+	Cluster               k8s.Cluster = k8s.Kubernetes
+	testKeplerImage       string
+	testKeplerRebootImage string
 )
 
 func TestMain(m *testing.M) {
@@ -26,6 +28,7 @@ func TestMain(m *testing.M) {
 	flag.StringVar(&controller.KeplerDeploymentNS, "deployment-namespace", controller.KeplerDeploymentNS,
 		"Namespace where kepler and its components are deployed.")
 	flag.StringVar(&testKeplerImage, "kepler-image", keplerImage, "Kepler image to use when running Internal tests")
+	flag.StringVar(&testKeplerRebootImage, "kepler-reboot-image", keplerRebootImage, "Kepler image to use when running PowerMonitorInternal tests")
 	flag.Parse()
 
 	if *openshift {
--- a/tests/e2e/main_test.go
+++ b/tests/e2e/main_test.go
@@ -15,12 +15,15 @@ import (
 const (
 	keplerImage       = `quay.io/sustainable_computing_io/kepler:release-0.7.12`
 	keplerRebootImage = `quay.io/sustainable_computing_io/kepler-reboot:v0.0.4`
+	ciTestVMEnvKey    = `powermonitor.sustainable.computing.io/test-env-vm`
 )
 
 var (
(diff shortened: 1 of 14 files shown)


KaiyiLiu1234:
e2e tests fully when I run make fresh and then run ./tests/run-e2e.sh --ci --no-upgrade. I am not sure why the e2e tests fail here. My guess is because it is using a different operator version? @vprashar2929 
Screenshot From 2025-04-30 16-52-32
Screenshot From 2025-04-30 16-53-17

tests/run-e2e.sh
```diff
@@ -206,10 +206,11 @@ run_e2e() {
 
 	log_events "$OPERATORS_NS" &
 	log_events "kepler-operator" &
+	log_events "power-monitor" &
 	watch_operator_errors "$error_log" &
 
 	local ret=0
-	run go test -v -failfast -timeout $TEST_TIMEOUT \
+	run go test -v -failfast -timeout $TEST_TIMEOUT -count=1 \
```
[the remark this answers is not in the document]
KaiyiLiu1234: -count=1 ensures that when running go test, it does not use cache. It will always run tests fresh. I guess it might not be necessary in the github action case because it start fresh. However, this will be more useful when running e2e tests on your own cluster.

tests/e2e/power_monitor_test.go
```diff
@@ -0,0 +1,204 @@
...
+	powermonitor := f.WaitUntilPowerMonitorCondition("power-monitor", v1alpha1.Available, v1alpha1.ConditionTrue)
+	assert.EqualValues(t, len(nodes), powermonitor.Status.Kepler.NumberAvailable)
+
+	f.DeletePowerMonitor("power-monitor")
+
+	f.AssertNoResourceExists(controller.PowerMonitorDeploymentNS, "", &corev1.Namespace{})
+	f.AssertNoResourceExists(ds.Name, ds.Namespace, &ds)
+}
+
+// below
```
sthaha, line 168:
```suggestion
```

sthaha (approved):
lgtm! minor comment.

sthaha (dismissed):

tests/run-e2e.sh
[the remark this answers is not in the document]
sthaha: @KaiyiLiu1234 lets handle this is a separate commit. IMHO not having -count=1 is useful and we can always pass -count=1 to e2e script if there is a need.

KaiyiLiu1234:
Issue discovered: Kepler deployed on e2e setup on github workflow does not have access to any rapl zones and the fake meter by default is disabled.

tests/e2e/main_test.go
```diff
@@ -13,19 +13,32 @@ import (
...
 	openshift := flag.Bool("openshift", true, "Indicate if tests are run aginast an OpenShift cluster.")
 	flag.StringVar(&controller.KeplerDeploymentNS, "deployment-namespace", controller.KeplerDeploymentNS,
 		"Namespace where kepler and its components are deployed.")
 	flag.StringVar(&testKeplerImage, "kepler-image", keplerImage, "Kepler image to use when running Internal tests")
+	flag.StringVar(&testKeplerRebootImage, "kepler-reboot-image", keplerRebootImage, "Kepler image to use when running PowerMonitorInternal tests")
+	flag.StringVar(&vmAnnotationKey, "vm-annotation-key", ciTestVMEnvKey, "VM Annotation Key set to enable vm test environment")
+	enableFakeMeter := "false"
+	if os.Getenv("ENABLE_VM_TEST") == "true" {
+		enableFakeMeter = "true"
+	}
```
sthaha, line 39: lets not make use of env vars when we can use `args`. My objection against ENV vars is that it modifies behavior silently

tests/run-e2e.sh
```diff
@@ -313,6 +320,7 @@ print_usage() {
 		⚙️ Options :
 		  -h|--help        show this help
 		  --ci             run in CI mode
+		  --enable-vm-test run tests in vm environment
```
sthaha, line 323: we can pass this flag to the e2e tests

sthaha (approved):

PR #482, opened 2025-04-29
chore: bump up kepler reboot

vprashar2929:
This commit bumps up Kepler reboot version to v0.0.5
It also copy the config package from Kepler so that we can use the config available in Operator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
 GIT_COMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
 
 KEPLER_VERSION ?=release-0.7.12
-KEPLER_REBOOT_VERSION ?=v0.0.4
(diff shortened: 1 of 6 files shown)


sthaha (changes requested):
Could you please copy the kepler/config package as well?

vprashar2929:
We should have some automated way to handle config. Like after release CI should notify Operator for version bump up and config updates(if any)

sthaha (commented):
LGTM! 
@KaiyiLiu1234 , could you please help validate this manually and merge it?

sthaha:
> We should have some automated way to handle config. Like after release CI should notify Operator for version bump up and config updates(if any)

No, rather operator will just import kepler/config

KaiyiLiu1234:
@sthaha  we already have a way to do this which is proven. Just need to decide when to add it again.

sthaha:
> @sthaha we already have a way to do this which is proven. Just need to decide when to add it again.

Sorry, didn't get that ... Would you mind elaborating?

sthaha (approved):

PR #484, opened 2025-05-01
ci: improve workflow execution

sthaha:
Changes include:
* Use large runners for workflows (cuts down time)
* Run both upgrade and e2e tests in parallel


KaiyiLiu1234 (approved):
lgtm

PR #488, opened 2025-05-02
ci: add dedicated go vet job to pr checks

vprashar2929:
This commit removes `go fmt` and `go vet` dependencies from `make build`. It adds a go vet job to PR check workflow, matching go fmt setup. It also updates the build-bundle and build-images jobs to run without any dependencies.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -8,12 +8,14 @@ jobs:
     runs-on: ubuntu-latest-8-cores
     steps:
       - uses: actions/checkout@main
+
       - uses: actions/setup-go@main
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-checks.yaml
```diff
@@ -23,18 +25,37 @@ jobs:
...
+      - uses: actions/checkout@main
+
+      - uses: actions/setup-go@main
+        with:
+          go-version-file: go.mod
+
+      - name: Install all tools
+        uses: ./.github/tools-cache
+
+      - name: make vet
```
sthaha, line 51: I think we should have a `lint` or `check` that runs all checks

sthaha (approved):

Issue #485, opened 2025-05-01
Improve e2e run script to accept sub-commands


sthaha:
Currently e2e/run script has 
```
	if $NO_UPGRADE; then
		deploy_and_run_e2e "$@" || return 1
	else
		run_bundle_upgrade || return 1
	fi
```

so --no-upgrade implies run e2e which isn't easy to understand. Lets make run script accept sub-commands to run so that the invocation is explicit as 
* ./tests/run.sh upgrade-test [args for upgrade test]
* ./test/run.sh e2e [args for e2e]

vprashar2929:
closing as #491 got merged

PR #491, opened 2025-05-03
test: update e2e script to accept commands

vprashar2929:
This commit updates the e2e script to now accept commands to run the e2e tests. Now we can run e2e tests with the following commands:
```
./tests/run.sh upgrade-test
./test/run.sh e2e [args for e2e]
```

Addresses: #485

### DIFF
--- a/tests/run-e2e.sh
+++ b/tests/run-e2e.sh
@@ -28,8 +28,8 @@ declare BUNDLE_IMG=""
 declare CI_MODE=false
 declare NO_DEPLOY=false
 declare NO_BUILDS=false
-declare NO_UPGRADE=false
 declare SHOW_USAGE=false
+declare COMMAND=""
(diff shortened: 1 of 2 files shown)


tests/run-e2e.sh
```diff
@@ -239,6 +240,39 @@ run_e2e() {
 declare -i ARGS_PARSED=0
 
 parse_args() {
```
sthaha, line 242: lets use the `cmd.sub-commmand`  trick to add sub-commands.  `main` should check for `cmd.$1` function and execute it.

sthaha (changes requested):
@vprashar2929 could you ensure tests are run when tests/ and config/ changes as well? Not just go files.

vprashar2929:
e2e tests always run

sthaha (approved):

PR #496, opened 2025-05-06
ci: make sure golang based jobs depend on the check-changes job

vprashar2929:
This commit adds the missing needs check in the golang based jobs.


KaiyiLiu1234 (approved):
lgtm

PR #509, opened 2025-05-13
tests(power-monitor): replace enable_vm_test env variable with command-line flag

KaiyiLiu1234:
Initially, enable_vm_test was populated using an enviornment variable. This was replaced with a command-line flag instead as it was more explicit.


sthaha (approved):

sthaha:
Will let @vprashar2929 take another look before merging

vprashar2929 (approved):

PR #510, opened 2025-05-15
feat: add support for custom config maps

vprashar2929:
This commit adds support for specifying the custom config maps to be used by the power monitor. Custom config map can be specified using the following config:
```
  kepler:
    config:
      logLevel: debug
      additionalConfigMaps:
        - name: my-conf
```
Operator expects the config map to be in the same namespace as the power monitor.


internal/controller/power_monitor_internal.go
```diff
@@ -60,21 +66,57 @@ func (r *PowerMonitorInternalReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
 
+// mapConfigMapToRequests returns the reconcile requests for power-monitor-internal objects for which an associated ConfigMap has changed
+func (r *PowerMonitorInternalReconciler) mapConfigMapToRequests(ctx context.Context, object client.Object) []reconcile.Request {
+	configMap, ok := object.(*corev1.ConfigMap)
+	if !ok {
+		return nil
+	}
+
+	pmis := v1alpha1.PowerMonitorInternalList{}
+	if err := r.List(ctx, &pmis); err != nil {
```
sthaha, line 96: Let use the `client` `r.Client`

internal/controller/power_monitor_internal.go
```diff
@@ -60,21 +66,57 @@ func (r *PowerMonitorInternalReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+	requests := []reconcile.Request{}
+	for _, pmi := range pmis.Items {
+		cms := pmi.Spec.Kepler.ConfigMaps
+		// Skip if no ConfigMaps are specified
+		if len(cms) == 0 {
+			continue
+		}
+
+		// Check if this ConfigMap is referenced
+		for _, cm := range cms {
```
sthaha, line 109: I have a feeling that this part can be simplified and that we don't have to iterate through all the `pmis` and all the configMap in the list.

vprashar2929: yeah right using indexes https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/client#FieldIndexer

internal/controller/power_monitor_internal.go
```diff
@@ -53,28 +61,67 @@ type PowerMonitorInternalReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *PowerMonitorInternalReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Set up indexer for PowerMonitorInternal based on referenced ConfigMaps
+	err := mgr.GetFieldIndexer().IndexField(context.TODO(), &v1alpha1.PowerMonitorInternal{}, configMapField, func(obj client.Object) []string {
```
sthaha, line 65:
```suggestion
	err := mgr.GetFieldIndexer().IndexField(context.Background(), &v1alpha1.PowerMonitorInternal{}, configMapField, func(obj client.Object) []string {
```

internal/controller/power_monitor_internal.go
```diff
@@ -53,28 +61,67 @@ type PowerMonitorInternalReconciler struct {
...
+		pmi := obj.(*v1alpha1.PowerMonitorInternal)
```
sthaha, line 66: Lets check if the conversion is okay and exit if it fails. (May be log it as well)

internal/controller/power_monitor_internal.go
```diff
@@ -37,6 +43,8 @@ type PowerMonitorInternalReconciler struct {
 	logger logr.Logger
 }
 
+const configMapField = ".spec.kepler.configMaps.name"
```
sthaha, line 46: isn't this `additionalConfigMaps` ?

sthaha (approved):

sthaha (approved):
Thanks a lot for this patch @vprashar2929 ❤️

Issue #508, opened 2025-05-13
Update reboot image for kepler operator tests


KaiyiLiu1234:
Currently, the reboot image used in kepler operator tests in version 0.0.4. We should update this to 0.0.5.

PR #511, opened 2025-05-15
test: use latest kepler release

vprashar2929:
This commit makes use of the latest kepler release when running Power Monitor Internal tests

Addresses #508


KaiyiLiu1234 (approved):
lgtm. close the correponding issue when you can

PR #512, opened 2025-05-16
workflow(xcrypto-check): add workflow to track x/crypto deps for operator

KaiyiLiu1234:
Workflow tracks golang.org/x/crypto deps introduced prior to PR, during PR, and any direct imports of golang.org/x/crypto in the operator. Any usage of x/crypto will result in a warning output in the logs but will not enforce a failure if x/crypto is discovered.

Note, x/crypto contains cryptographic modules that are unacceptable for fips compliance. It is important to keep track of x/crypto usage to aid in manual audits.


KaiyiLiu1234:
Workflow added to Kepler Reboot here: https://github.com/sustainable-computing-io/kepler/pull/2069

KaiyiLiu1234:
Sample CI Run when x/crypto dependencies get introduced: https://github.com/KaiyiLiu1234/kepler/actions/runs/15060128577/job/42333613680?pr=5. Although test was performed on Kepler, both the workflows in kepler and operator are the same.

KaiyiLiu1234:
Note pre-commit fails due to this:
```
In hack/reports/print-xcrypto-report.sh line 28:
        echo "$new_deps" | sed 's/^/ 🔹 /'
        ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.


In hack/reports/print-xcrypto-report.sh line 42:
    echo "$occurrences" | sed 's/^/ 🔹 /'
    ^-- SC2001 (style): See if you can use ${variable//search/replace} instead.
```
I would prefer to continue to use sed as it is far cleaner. I can rebase to disable shellcheck for these instances. thoughts?

hack/reports/print-xcrypto-report.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+echo "=== Dependency Report ==="
+echo ""
+
+# Check dependencies prior to PR
+echo "ℹ️ Current Kepler dependencies on golang.org/x/crypto:"
```
vprashar2929, line 8:
```suggestion
echo "ℹ️ Current Kepler Operator dependencies on golang.org/x/crypto:"
```

KaiyiLiu1234: Added change!

vprashar2929:
@KaiyiLiu1234 How about we just check for xcrypto on every pull request instead against base? We are only interested in finding whether the current PR introduces any xcrypto changes or not. Also, it will allow us to run the reporting script locally before even creating a PR as well. Thoughts?

KaiyiLiu1234:
@vprashar2929 change has been made to the kepler repo instead to address your concern. However, I believe there is a permissions error when it comes to have the workflow make a comment. Is there a way write access can be provided (i do not have the permissions to do this)

KaiyiLiu1234:
@vprashar2929 @sthaha  this should be good to go for x crypto checks. Note this run however: https://github.com/sustainable-computing-io/kepler-operator/actions/runs/16240626190/job/45856567081. I am  not sure why it fails because after i re run, it succeeds. We probably should look into why this is happening.

PR #524, opened 2025-05-29
feat: secure kepler's /metrics endpoint with kube rbac proxy

KaiyiLiu1234:
Power Monitor CR has been updated with a Security Mode to protect kepler's /metrics endpoint. When Mode is switched to 'rbac', the power monitor daemonset will be updated with kube rbac proxy sidecar which will enforce which service accounts can access /metrics based on static configuration and clusterrole/role rules. TLS is also enabled using kube rbac proxy and certs produced by openshift. Power Monitor API has been updated to include 'allowedServiceAccounts' which will allow users to request certain service accounts to have access.


KaiyiLiu1234:
I will rebase once we confirm this implementation is acceptable

pkg/components/power-monitor/deployment.go
```diff
@@ -436,6 +649,63 @@ func newPowerMonitorContainer(pmi *v1alpha1.PowerMonitorInternal) corev1.Contain
 	}
 }
 
+func newKubeRBACProxyContainer(pmi *v1alpha1.PowerMonitorInternal) corev1.Container {
+	return corev1.Container{
+		Name:  KubeRBACProxyContainerName,
+		Image: "quay.io/brancz/kube-rbac-proxy:v0.19.0",
```
vimalk78, line 655: this shiould come from CSV

vimalk78 (changes requested):

PR #528, opened 2025-06-02
ci: specify codecov secret to the workflow

vprashar2929:
This commit fixes the issue where the codecov calling workflow needs secret to be passed explicitly.

It also renames the publish-images workflow to push workflow for better clarity.


vprashar2929:
fixes this CI failure: https://github.com/sustainable-computing-io/kepler-operator/actions/runs/15381350756

sthaha (approved):

PR #530, opened 2025-06-03
feat(pod): Operator support Pod power

vimalk78:
This PR requires changes from https://github.com/sustainable-computing-io/kepler/pull/2127 

Updated config 
Updated deployment spec
Updated tests

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -71,6 +71,12 @@ type (
 		Pprof PprofDebug `yaml:"pprof"`
 	}
 
+	Kube struct {
+		Enabled *bool  `yaml:"enabled"`
+		Config  string `yaml:"config"`
+		Node    string `yaml:"nodeName"`
+	}
+
 	Config struct {
 		Log      Log      `yaml:"log"`
 		Host     Host     `yaml:"host"`
@@ -80,6 +86,8 @@ type (
 		Web      Web      `yaml:"web"`
 		Debug    Debug    `yaml:"debug"`
 		Dev      Dev      `yaml:"dev"` // WARN: do not expose dev settings as flags
+
+		Kube Kube `yaml:"kube"`
 	}
 )
 
@@ -87,6 +95,7 @@ type SkipValidation int
 
 const (
 	SkipHostValidation SkipValidation = 1
+	SkipKubeValidation SkipValidation = 2
 )
 
 const (
@@ -113,6 +122,11 @@ const (
 	ExporterPrometheusEnabledFlag = "exporter.prometheus"
 	// NOTE: not a flag
 	ExporterPrometheusDebugCollectors = "exporter.prometheus.debug-collectors"
+
+	// kubernetes flags
+	KubernetesFlag   = "kube.enable"
+	KubeConfigFlag   = "kube.config"
+	KubeNodeNameFlag = "kube.node-name"
 
 // WARN:  dev settings shouldn't be exposed as flags as flags are intended for end users
 )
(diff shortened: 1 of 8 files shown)


internal/config/config_test.go
```diff
@@ -473,7 +500,6 @@ func TestWebConfig(t *testing.T) {
 	t.Run("valid web config", func(t *testing.T) {
 		tempWebConfig, err := os.CreateTemp("", "temp_*web.yml")
 		assert.NoError(t, err, "cannot create temp file")
-		defer os.Remove(tempWebConfig.Name())
```
sthaha, line 476: Lets add it to cleanup instead `t.Cleanup()` so that if test fails in between the file is cleanup.

internal/config/config.go
```diff
@@ -71,6 +71,12 @@ type (
 		Pprof PprofDebug `yaml:"pprof"`
 	}
 
+	Kube struct {
+		Enabled *bool  `yaml:"enabled"`
+		Config  string `yaml:"config"`
+		Node    string `yaml:"nodeName"`
```
sthaha, line 77:
```suggestion
		NodeName    string
```

Storing the NodeName in config yaml is a double edged sword .. It implies that all keplers on all nodes will watch the Node in the config.yaml unless --kube.node is passed.

internal/config/config.go
```diff
@@ -71,6 +71,12 @@ type (
 		Pprof PprofDebug `yaml:"pprof"`
 	}
 
+	Kube struct {
+		Enabled *bool  `yaml:"enabled"`
+		Config  string `yaml:"config"`
```
sthaha, line 76: how about we call it ConfigFile?

internal/config/config.go
```diff
@@ -343,6 +385,19 @@ func (c *Config) Validate(skips ...SkipValidation) error {
 			errs = append(errs, fmt.Sprintf("invalid monitor staleness: %s can't be negative", c.Monitor.Staleness))
 		}
 	}
+	{ // Kubernetes
+		if c.Kube.Config != "" && !ptr.Deref(c.Kube.Enabled, false) {
+			errs = append(errs, fmt.Sprintf("%s supplied but %s set to false", KubeConfigFlag, KubernetesFlag))
+		}
+		if c.Kube.Node != "" && !ptr.Deref(c.Kube.Enabled, false) {
```
sthaha, line 392:
```suggestion
		if c.Kube.Node != "" && !ptr.Deref(c.Kube.Enabled, false) {
```
shouldn't we insist that this isn't empty? other wise it will watch all nodes - right?

internal/config/config.go
vimalk78: `--kube.node-name` is mandatory in kepler

internal/config/config.go
vimalk78: changed

pkg/components/power-monitor/deployment.go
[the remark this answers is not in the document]
vimalk78: seems to be working fine now

pkg/components/power-monitor/deployment.go
```diff
@@ -415,9 +415,15 @@ func newPowerMonitorContainer(pmi *v1alpha1.PowerMonitorInternal) corev1.Contain
 		Name:            pmi.DaemonsetName(),
 		SecurityContext: &corev1.SecurityContext{Privileged: ptr.To(true)},
 		Image:           deployment.Image,
```
sthaha, line 417: @vimalk78 this needs the new image - right? Could you please update the image to 0.0.8 ?

vprashar2929 (approved):

PR #533, opened 2025-06-07
ci: ensure only powermonitor tests are executed

vprashar2929:
This commit ensures that only powermonitor tests are executed in the CI. It introduces `-skip-kepler-tests` flag that when set, skips all kepler tests


sthaha (changes requested):
Lets not do that since we should all now ensure that the test starts with PowerMonitor. Lets instead use `t.Skip()` on all Kepler tests if `--skip.kepler` is passed. (see test main).

PR #535, opened 2025-06-11
chore: make kepler CR deprecated

vprashar2929:
This commit sets the Kepler CR status as deprecated. It also removes the existing reconcilers ensuring that when Kepler CR is created it only reconciles openshift dashboards and sets the status of CR as degraded
It also adds a new conditions field to Kepler status which ensures that deprecation status is reflected properly on OpenShift console.


vprashar2929:
OpenShift console now shows the Kepler deprecation status:

Screenshot 2025-06-12 at 8 29 00 PM

sthaha (commented):
lgtm! lets wait to record the demo before merging this 🙏

PR #555, opened 2025-07-07
chore: add CPU Info panel to overview dashboard

vprashar2929:
This commit updates the overview dashboard of PowerMonitor CR:
* Removes `All` Nodes from the dropdown
* Update Current Power Consumption to remove nodes from the promql query
* Move the Current Power Consumption inside `Cluster Wide Consumption` row
* Adds a new `Node Info` row which shows the CPU Info table with Node Name, Model Name and Cores count
* Update the `Node Usage` title to `Node Consumption`


vprashar2929:
Dashboard Screenshot

Metrics-dashboards-07-07-2025_03_40_PM

PR #563, opened 2025-07-09
feat(spec): add user-configurable metrics level, staleness, sample rate, and max terminated fields to PowerMonitor spec

vimalk78:
- Add MetricsLevel field to control which metrics levels to export (node, process, container, vm, pod)
  - Add Staleness field to control how long to wait before considering calculated power values as stale (default: 500ms)
  - Add SampleRate field to control the interval for monitoring resources (default: 5s)
  - Add MaxTerminated field to control terminated workload tracking behavior (default: 500)
  - Update PowerMonitor and PowerMonitorInternal CRDs with new fields
  - Add tests


api/v1alpha1/power_monitor_internal_types.go
```diff
@@ -39,6 +41,29 @@ type PowerMonitorInternalKeplerConfigSpec struct {
 	// +optional
 	// +listType=atomic
 	AdditionalConfigMaps []ConfigMapRef `json:"additionalConfigMaps,omitempty"`
+
+	// MetricsLevel specifies which metrics levels to export
+	// Valid values are combinations of: node, process, container, vm, pod
+	// +optional
+	MetricsLevel config.Level `json:"metricsLevel,omitempty"`
```
sthaha, line 48: Can we rename this to `MetricLevels` to indicate plural. (we should fix kepler.config) and make use of `// +kubebuilder:validation:items:Enum` and make this a string list?

How does the current validation work?

api/v1alpha1/power_monitor_types.go
```diff
@@ -28,6 +32,18 @@ type PowerMonitorKeplerDeploymentSecuritySpec struct {
...
 }
 
+// MetricsLevelDefault represents the default metric levels for PowerMonitor (node and pod)
+const MetricsLevelDefault = config.MetricsLevelNode | config.MetricsLevelPod
+
+// StalenessDefault represents the default staleness duration for PowerMonitor (500ms)
+const StalenessDefault = 500 * time.Millisecond
+
+// SampleRateDefault represents the default sample rate for PowerMonitor (5s)
+const SampleRateDefault = 5 * time.Second
```
sthaha, line 42: do we need this? isn't it covered by default?

api/v1alpha1/power_monitor_types.go
```diff
@@ -28,6 +32,18 @@ type PowerMonitorKeplerDeploymentSecuritySpec struct {
 	AllowedSANames []string `json:"allowedSANames,omitempty"`
 }
 
+// MetricsLevelDefault represents the default metric levels for PowerMonitor (node and pod)
+const MetricsLevelDefault = config.MetricsLevelNode | config.MetricsLevelPod
+
+// StalenessDefault represents the default staleness duration for PowerMonitor (500ms)
+const StalenessDefault = 500 * time.Millisecond
```
sthaha, line 39: Doesn't the default already set this?

internal/config/config_test.go
```diff
@@ -804,3 +920,967 @@ exporter:
...
+			args:          []string{"--metrics", "invalid"},
+			expectedLevel: MetricsLevelAll, // Should remain at default
+			expectError:   true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a kingpin application for testing
+			app := kingpin.New("test", "test application")
```
sthaha, line 1208: Does the test fail? 🤔

sthaha (approved):

Issue #556, opened 2025-07-07
In case of k8s user waits for 1m for CR status to update


vprashar2929:
When PowerMonitor is deployed on K8s with security mode as rbac users have to wait for 1m for the CR status to update which can be annoying. We can reduce the wait time for tls secret to be available in power-monitor namespace

Issue #557, opened 2025-07-07
Timeout for rbac checks should change depending on cluster


KaiyiLiu1234:
In a kubernetes cluster, the timeout should be much shorter than 1 minute (current timeout) then for an openshift cluster. On openshift cluster, relevant certs are expected to be created by relevant operators and waiting may be needed. In kubernetes cluster, relevant certs must be made manually by the user, so the timeout for missing certs should be shorter.

PR #565, opened 2025-07-10
chore: change timeout depending on cluster type

KaiyiLiu1234:
When cluster type is OpenShift, the timeout is changed to 60 seconds. When cluster type is K8s, the timeout is changed to 6 seconds.


KaiyiLiu1234:
@vprashar2929 Can you please manually validate this?

KaiyiLiu1234:
Fixes #556 #557

sthaha (approved):

PR #566, opened 2025-07-10
chore: ensure that Kepler listen on 127.0.0.1

vprashar2929:
This commit ensures that Kepler listens on `127.0.0.1:28282` instead of `0.0.0.0:28282`.


KaiyiLiu1234:
@vprashar2929 I believe this changed created a problem with the liveness check. If you check the liveness probe for rbac tests, it now says unhealthy.

vprashar2929:
@KaiyiLiu1234 Good find! 🙇 I missed that completely. Will have the fix for that

PR #567, opened 2025-07-10
chore: add example for security setting in PowerMonitor CR

vprashar2929:
This commit updates the PowerMonitor CR example to include security settings with default values

### DIFF
--- a/config/samples/kepler.system_v1alpha1_powermonitor.yaml
+++ b/config/samples/kepler.system_v1alpha1_powermonitor.yaml
@@ -10,3 +10,6 @@ spec:
   kepler:
     config:
       logLevel: info
+    deployment:
+      security:
+        mode: none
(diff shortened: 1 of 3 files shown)


vprashar2929:
OpenShift Console:

Screenshot 2025-07-10 at 2 43 27 PM

config/samples/kepler.system_v1alpha1_powermonitor.yaml
```diff
@@ -10,3 +10,8 @@ spec:
   kepler:
     config:
       logLevel: info
+    deployment:
+      security:
+        mode: none
+        allowedSANames:
```
KaiyiLiu1234, line 16: I don't think you need to include this because if this is not included, the default will be added depending on cluster type already.

sthaha (approved):

Issue #558, opened 2025-07-07
Errors in deployment.go should not be ignored and should be propogated


KaiyiLiu1234:
Issue Reference: https://github.com/sustainable-computing-io/kepler-operator/pull/554#discussion_r2188972576

KaiyiLiu1234:
Another occurrence of an error being ignored in deployment.go: https://github.com/sustainable-computing-io/kepler-operator/blob/v1alpha1/pkg/components/power-monitor/deployment.go#L209

PR #570, opened 2025-07-11
chore: add error propagation to deployment.go

KaiyiLiu1234:
For builders that can return an error  like KeplerConfig and NewPowerMonitorKubeRBACProxyConfig, errors from these functions are now returned to the reconcilers to be handled gracefully.


KaiyiLiu1234:
@vprashar2929 I do not believe I can do anything about codecov as it checks security.go and power-monitor.go (which is tested in e2e but not recognized by codecov). Also, we should rename power-monitor.go as in actuality it is used to reconicle the configmap

KaiyiLiu1234:
Fixes #558. Note, that in deployment_test.go, propagating the error actually revealed an error in the TestNewPowerMonitorConfigMap test function (loglevel was not set in the default case). This was fixed in this patch.

sthaha:
@KaiyiLiu1234 , LGTM but can we please improve the test coverage ?

PR #571, opened 2025-07-14
fix: resolve liveness probe failure in RBAC mode

vprashar2929:
This commit resolves the issue where the liveness probe fails in RBAC mode by replacing the HTTP GET liveness probe with exec probe.

In case of RBAC mode, container listens on 127.0.0.1:28282 and HTTP GET liveness probe runs from kubelet. Kubelet cannot reach to pod's localhost causing the probe failure.

Exec probe runs inside container and can reach host in both the modes

Addresses: https://github.com/sustainable-computing-io/kepler-operator/pull/566#issuecomment-3067169985


KaiyiLiu1234:
Issue appears to be resolved.

KaiyiLiu1234 (approved):

PR #572, opened 2025-07-15
test: ensure RBAC e2e can run on OpenShift

vprashar2929:
This commit updates the e2e test for RBAC ensuring that it can run on OpenShift. Existing e2e tests were specific to Kubernetes hence deploys prerequisites for OpenShift on Kubernetes like cert-manager certs, etc. which are not required when running tests against an OpenShift cluster.

In case of OpenShift, it creates the ConfigMap in `testNs` namespace with the annotation `service.beta.openshift.io/inject-cabundle": "true"` OpenShift automatically detects this annotation and injects the service CA certs leveraging OpenShift's automatic CA injection mechanism.

### DIFF
--- a/tests/e2e/power_monitor_internal_test.go
+++ b/tests/e2e/power_monitor_internal_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sustainable.computing.io/kepler-operator/api/v1alpha1"
 	"github.com/sustainable.computing.io/kepler-operator/internal/controller"
 	powermonitor "github.com/sustainable.computing.io/kepler-operator/pkg/components/power-monitor"
+	"github.com/sustainable.computing.io/kepler-operator/pkg/utils/k8s"
 	"github.com/sustainable.computing.io/kepler-operator/pkg/utils/test"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


KaiyiLiu1234:
@vprashar2929  I dont see why we need this: CreateCAConfigMapFromOpenshiftServiceCert. We should instead wait for the operator to have Reconciled: true and Available: true. The operator handles creating this and waiting for this resource on openshift enviornment. Second, CreateCurlPowerMonitorTestSuiteForOpenShift seems to be a duplicate. How you run the CurlPowerMonitorTestSuite should be the same in kubernetes and openshift I believe (Correct me if i am wrong).

tests/e2e/power_monitor_internal_test.go
```diff
@@ -124,40 +125,125 @@ func TestPowerMonitorInternal_RBAC_Reconciliation(t *testing.T) {
...
+			caCertName,
+			caCertSecretName,
+			pmIssuerName,
+			tlsCertSecretName,
+			tlsCertSecretName,
+		)
+		caCertSource = caCertSecretName
+	} else {
+		f.WaitUntilPowerMonitorInternalCondition(pmi.Name, v1alpha1.Reconciled, v1alpha1.ConditionTrue)
+		f.WaitForOpenshiftCerts(pmi.Name, testNs, tlsCertSecretName)
```
KaiyiLiu1234, line 151: So it looks like you are waiting for this, which is good. I would rather use Cluster == k8s.OpenShift rather than use else. Because the certs being added are only automatically added by openshift as far as we know. So in other clusters like kubernetes we should expect to manually add these "openshift" certs.

KaiyiLiu1234 (approved):
lgtm

PR #576, opened 2025-07-17
chore: move utils outside e2e

vprashar2929:
This commit moves the utils folder from tests/e2e to tests.
This also reverts the source code go.mod file back to the original one before cert-manager was added.

Ref: https://github.com/sustainable-computing-io/kepler-operator/blob/chore-version-0.18.0/go.mod


KaiyiLiu1234:
@vprashar2929 looks great! How did you resolve the snyk issue? Was it by dropping the version of apiserver from 0.31.1 to 0.31.0?

vprashar2929:
@sthaha Definitely a must to have

PR #589, opened 2025-07-24
chore: remove SYS_ADMIN cap from SCC

vprashar2929:
This commit removes the SYS_ADMIN capability from the SCC as Kepler doesn't need it.


pkg/components/power-monitor/deployment.go
```diff
@@ -338,7 +338,7 @@ func NewPowerMonitorSCC(d components.Detail, pmi *v1alpha1.PowerMonitorInternal)
 		AllowHostNetwork:         false,
 		AllowHostPID:             true,
```
sthaha, line 339: Do we need host PID? actually, do we need SCC?

sthaha (changes requested):
Please reconsider if we really need SCC.

PR #592, opened 2025-07-29
chore: update deploy grafana script

vprashar2929:
This commit updates the deploy grafana script according to new PowerMonitor CR

It also updates the dashboard to match with what we have in the OpenShift dashboard

### DIFF
--- a/hack/dashboard/assets/kepler/dashboard.json
+++ b/hack/dashboard/assets/kepler/dashboard.json
@@ -18,12 +18,11 @@
       }
     ]
   },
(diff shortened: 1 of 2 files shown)


sthaha (approved):

hack/dashboard/assets/kepler/dashboard.json
sthaha, line 1: @vprashar2929 can we add a "job_name" variable which defaults to power-monitor, so that we can easily test this? Even better would be a drop down of `kepler_` jobs.

PR #595, opened 2025-08-04
ci: set go cache to false

vprashar2929:
This commit disables go cache in the workflow ensuring a clean environment for each run. This avoids conflicts with existing files like those causing `tar` errors


vprashar2929:
Ref: https://github.com/sustainable-computing-io/kepler-operator/actions/runs/16715188509/job/47307188688?pr=593#step:7:97

PR #612, opened 2025-10-06
docs: add user guide for additionalConfigMaps

vprashar2929:


docs/user-guides/configuring-kepler.md
```diff
@@ -0,0 +1,255 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-kepler-config
+  namespace: power-monitor
+data:
+  config.yaml: |
+    metrics:
+      export:
+        interval: 10s
```
sthaha, line 35: Is this valid?

docs/user-guides/configuring-kepler.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kepler-monitor-config
+  namespace: power-monitor
+data:
+  config.yaml: |
+    monitor:
+      interval: 10s        # Monitor resources every 10 seconds (default: 5s)
```
sthaha, line 103: these can't overridden by additionalConfig since `spec.kepler.config.interval` should take precedence over any additional config.

PR #616, opened 2025-10-08
ci(release): push helm chart before creating pull request

vprashar2929:


sthaha (dismissed):

.github/workflows/release.yaml
```diff
@@ -82,6 +82,11 @@ jobs:
           make operator-push  bundle-push
         env:
           IMG_BASE: ${{ vars.IMG_BASE }}
+      - name: Push Helm Chart to OCI registry
+        shell: bash
+        run: |
+          VERSION=${{ github.event.inputs.tag }}
+          helm push helm-releases/kepler-operator-helm-${VERSION}.tgz oci://quay.io/sustainable_computing_io/charts
```
sthaha, line 89: we should remove the helm-releases folder once the chart is pushed and not include the tarball in the release

sthaha: In hindsight we should have used tmp/helm-releases for generating the chart.

sthaha (approved):
LGTM!

PR #647, opened 2025-11-25
chore: update script to validate helm

vprashar2929:
This commit updates the validate script of helm to properly handle the path to the tools necessary to be installed before running the validation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -336,7 +336,7 @@ helm-sync-crds: ## Sync CRDs from config/crd/bases to Helm chart
 	@echo "✅ CRDs synced to Helm chart"
 
 .PHONY: helm-validate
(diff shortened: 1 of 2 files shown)


vprashar2929:
Also addresses the issue where the [CI](https://github.com/sustainable-computing-io/kepler-operator/actions/runs/19660901786/job/56308352324#step:5:8) fails when running helm-validate job as the GH runner already contains the [kustomize](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#tools) binary and the validate script expects it to be under `$PROJECT_ROOT/tmp/bin`

vimalk78 (approved):
lgtm

PR #651, opened 2025-12-03
fix(security): replace privileged with CAP_SYS_PTRACE

vprashar2929:
Apply principle of least privilege by dropping all capabilities except CAP_SYS_PTRACE which Kepler needs to read /proc/<pid> Also enforce readonly root filesystem and prevent privilege escalation

Security improvements:
- Set privileged: false (no full root capabilities)
- Drop ALL capabilities, add only SYS_PTRACE
- Set allowPrivilegeEscalation: false
- Set readOnlyRootFilesystem: true
- Updated OpenShift SCC with same constraints


vimalk78:
pls mention if this fix is tested on which openshift version

vimalk78 (approved):

PR #653, opened 2025-12-22
fix: remove blocking waits during deletion

vprashar2929:
Replace blocking wait.PollUntilContextTimeout with non-blocking check-and-requeuepattern in Deleter reconciler. This reduces PowerMonitor CR deletion time from5+ minutes to under 30 seconds by allowing Kubernetes GC to run in parallelinstead of sequentially polling for each resource deletion.

Changes:
- Refactor Deleter to use check-and-requeue instead of blocking polls
- Remove WaitTimeout field from Deleter (no longer needed)
- Update Deleter tests for non-blocking behavior
- Remove 2-minute namespace deletion wait in PowerMonitorInternal cleanup

### DIFF
--- a/internal/controller/power_monitor.go
+++ b/internal/controller/power_monitor.go
@@ -202,6 +202,8 @@ func (r PowerMonitorReconciler) reconcilersForPowerMonitor(pm *v1alpha1.PowerMon
 		detail = components.Full
 	}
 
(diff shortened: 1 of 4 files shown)


vimalk78:
doesn't having blocking behavior guarantees operation ordering. with non-blocking need to manage state more carefully

